PR #16, opened 2023-04-21
[WIP] Commit the alpha-version image deconvolution classes

hiyoneda:
I commit the initial version of image deconvolution classes. Obviously, I still need to connect them with the data IO class and image response class, but I do this to share my thoughts about the image deconvolution class to the team.


israelmcmc:
@hiyoneda Great! I think it's good to share code at an early stage. Would it be possible to add a notebook with an example on how to run it? Either in the docs folder here or in intro_cosipy.

hiyoneda:
Yes, I will share the notebook in intro_cosipy.

PR #22, opened 2023-05-03
Add example notebook for image deconvolution

tsiegert:


hiyoneda:
It works. Note that after updating COSIpy, I needed to modify image_deconvolution.py to remove "import HealpixAxis".

PR #35, opened 2023-05-05
Class to manipulate the orientation file:

saurabhmittal23:
- time of the pointings as a time object
 - spacecraft attitude as scoords attitude object
 - time difference between adjacent observations


israelmcmc:
Note: __init__ files are missing

PR #42, opened 2023-05-08
add ts_map in cosipy

PbU-Jason:
I have migrated my previous TS map code from Jupyter notebook in intro_cosipy to .py format in cosipy. The logical structure is as follows,

1.  In "COSILikeForTSMap.py", I created "class COSILikeForTSMap" to inherit "class COSILike" in "cosipy.threeml.COSILike", but override the "set_model" function, because in the TS map COSILikeForTSMap needs to change RA and Dec many times.
2. In "TSMap.py", I created "class TSMap". "link_model_all_plugins" function use dr, data, bkg, sc_orientation, piv and index to instantiate plugin, create model and link them together to wait for the fit. "ts_fitting" function allows the user to take care of source and bkg to generate the TS grid. There are some remaining functions that mainly help the user to visualize, save or load the final result.


ckierans:
I checked the TSmap pull request and it works with no issues.

israelmcmc:
Hi @PbU-Jason, thanks! We think it is almost ready to merge (you can add more feature later), just change these couple of things before:
- Use the Attitude and Spacecraft classes from the scoords library, not from cosipy.coordinates. In the latest version cosipy, these are no longer there. See #43 
- Instead of adding a class in cosipy/ts_map/COSILikeForTSMap.py, apply your modification on top of the existing COSILike class here: https://github.com/cositools/cosipy/blob/main/cosipy/threeml/COSILike.py
- Remove the .DS_Store files from this branch

PbU-Jason:
> Hi @PbU-Jason, thanks! We think it is almost ready to merge (you can add more feature later), just change these couple of things before:
> 
> * Use the Attitude and Spacecraft classes from the scoords library, not from cosipy.coordinates. In the latest version cosipy, these are no longer there. See [Use scoords for Attitude and SpacecraftFrame #43](https://github.com/cositools/cosipy/pull/43)
> * Instead of adding a class in cosipy/ts_map/COSILikeForTSMap.py, apply your modification on top of the existing COSILike class here: https://github.com/cositools/cosipy/blob/main/cosipy/threeml/COSILike.py
> * Remove the .DS_Store files from this branch

Thanks for the guidance. I've fixed above issues.

ckierans:
Checked out the new commit and things work now with scoords and with COSILike. Ready to merge.

PR #58, opened 2023-05-11
Update on the image deconvolution in response to the new orientation handling

hiyoneda:
- created a new class, "DataLoader", which calculates the dwell time map using the orientation file
- removed the class "DataIO_dummy"
- changed the input parameter file format for the image deconvolution
    - the definition of the energy bands in the model map
    -  the definition of the initial value in each energy band in the model map
- added functionality in order to save the output at the end of each iteration


hiyoneda:
I added the notebook for the image analysis of the 2-sec GRB data. See cosipy/docs/tutorials/image_deconvolution_test_2sGRB_20iteration_w_saving_results_20230511.ipynb.

The parameter file of the image deconvolution is cosipy/test_data/imagedeconvolution_parfile_itr10_20230511.yml.

Note that the input hdf5 file is not uploaded. So the file path should be edited by a user.

hiyoneda:
I modified DataLoader. Now we can use the event/background binned data, response and orientation objects directly.

israelmcmc:
I tested this in the GRB master notebook and it works. Merging.

PR #67, opened 2023-05-18
Master GRB notebook

israelmcmc:
Still need to clean it up and comment it out....


israelmcmc:
This is deprecated with DC2

PR #68, opened 2023-06-06
updated dataIO

ckarwin:
Major updates:

1. I fixed the issue with psi and chi local and galactic. I had a sign error in the vector from  the second interaction to the first interaction. Everything is now in good agreement after fixing this. Specifically, I've checked psi and chi local and galactic using the grb from mini-DC2. Additionally, I've verified that all values are in good agreement with the MEGAlib calculations. 

2. I've added the first test script for dataIO. It compares psi and chi to the MEGAlib calculations. This comparison requires a local MEGAlib installation, so it will only be ran by certain users. The code should now account for this and not throw errors for users without MEGAlib -- I've wrapped the corresponding imports in a try statement. I will add more tests soon. 

3. I've added another method to easily plot slices of psi and chi, in order to view a slice of the Compton cone. 

4. When binning the data you can now choose between local and Galactic coordinates, for psi and chi. The default is Galactic.


ckarwin:
I've updated the MEGAlib comparison test, as recommended by @israelmcmc. The test now reads in the MEGAlib calculations from a file instead of using MEGAlib to calculate them on the fly. The file is kept in cosipy/test_data. Additionally, I have added all parameters to the test; the last version was only testing psi and chi. 

For now I am keeping the old TRA reader class in dataIO. This should not cause any issues, but please let me know if any problem arises. I think it's important to keep this for the time being, as it will make future testing much more easier.

PR #69, opened 2023-06-22
Update FullDetectorResponse.py for reading rsp file

GallegoSav:
I modified the `open` class method in order to check if the response filename is `.h5` or `.rsp`. Then it will call the method `_open_h5` (the current used) or `_open_rsp` wich read the rsp file, create a `.h5` file with the good structure.

### DIFF
--- a/cosipy/response/FullDetectorResponse.py
+++ b/cosipy/response/FullDetectorResponse.py
@@ -1,36 +1,26 @@
+from .PointSourceResponse import PointSourceResponse
+from .DetectorResponse import DetectorResponse
+from astromodels.core.model_parser import ModelParser
+import matplotlib.pyplot as plt
+from astropy.time import Time
+from astropy.coordinates import SkyCoord
+from astropy.units import Quantity
+import astropy.units as u
+from sparse import COO
+from pathlib import Path
+import numpy as np
+import mhealpy as hp
+from mhealpy import HealpixBase, HealpixMap
+from cosipy.config import Configurator
+from scoords import SpacecraftFrame
+from histpy import Histogram, Axes, Axis, HealpixAxis
+import h5py as h5
+import os
+import textwrap
+import argparse
 import logging
 logger = logging.getLogger(__name__)
 
-import argparse
-import textwrap
-
-import h5py as h5
-
-from histpy import Histogram, Axes, Axis, HealpixAxis
-
-from scoords import SpacecraftFrame
-from cosipy.config import Configurator
-
-from mhealpy import HealpixBase, HealpixMap
-import mhealpy as hp
-
-import numpy as np
-
-from pathlib import Path
-
-from sparse import COO
-
-import astropy.units as u
-from astropy.units import Quantity
-from astropy.coordinates import SkyCoord
-from astropy.time import Time
-
-import matplotlib.pyplot as plt
-
-from astromodels.core.model_parser import ModelParser
-
-from .DetectorResponse import DetectorResponse
-from .PointSourceResponse import PointSourceResponse
 
 class FullDetectorResponse(HealpixBase):
     """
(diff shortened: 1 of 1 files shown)


israelmcmc:
@GallegoSav Do you have an example on how to run this? Maybe your copy of the DetectorResponse.ipynb that you used to open PowerlawContinuum_index2.Medium.binnedimaging.imagingresponse.rsp ?

cosipy/response/FullDetectorResponse.py
```diff
@@ -39,63 +31,423 @@ class FullDetectorResponse(HealpixBase):
...
+
+        # Full-sky?
+        if not single_pixel:
+
+            # Assumming all FISBEL pixels have the same area
+            nperchannel /= dr.axes['NuLambda'].nbins
+
+        # Area
+        counts2area = area_sim / nperchannel
+        dr_area = dr * dr.expand_dims(counts2area, 'Ei')
```
snpike, line 314: When I run this with mono normalization, I get an error saying that the dimensions of dr['Ei'] and counts2area don't match up. I think this is because counts2area is a float.

cosipy/response/FullDetectorResponse.py
```diff
@@ -39,63 +31,423 @@ class FullDetectorResponse(HealpixBase):
...
+            nperchannel_norm[ecenters > emax] = 0
+
+
+        elif norm =="Linear" :
+            print("normalisation : linear with energy range [{0}-{1}]".format(emin,emax))
+            nperchannel_norm = ewidth / (emax-emin)
+            
+        elif norm=="Mono" :
+            print("normalisation : mono")
+            nperchannel_norm = 1.
```
snpike, line 302:
```suggestion
            nperchannel_norm = np.ones_like(dr.axes['Ei'])
```

GallegoSav: Thanks for pointig this out. The solution was to do `nperchannel_norm = [1.]`

israelmcmc:
I know that we're still working on this, but I run the line profiles to know that takes so long. 85% of the time is on these lines, as expected:
```
File: /Users/imartin5/software/cosipy/cosipy/response/FullDetectorResponse.py
Function: _open_rsp at line 128

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   259                                                   # read the rsp file and get the bin number and counts
   260         1     441000.0 441000.0      0.0          with gzip.open(filename, "rt") as file:
   261                                                        
   262         1    3637000.0 3637000.0      0.0              progress_bar = tqdm(file, total=nlines, desc="Progress", unit="line")
   263                                           
   264  26362000 45771167000.0   1736.3     15.7              for line in progress_bar:
   265                                           
   266                                                           
   267  26362000 15865711000.0    601.8      5.4                  line = line.split()
   268                                           
   269  26361990 12038971000.0    456.7      4.1                  if len(line) == 0:
   270        10       1000.0    100.0      0.0                      continue
   271                                           
   272  26361990 10202978000.0    387.0      3.5                  key = line[0]
   273                                           
   274  26361933 9335179000.0    354.1      3.2                  if key == 'RD':
   275                                           
   276  26361933 53721280000.0   2037.8     18.4                      b = np.array(line[1:-1], dtype=int)
   277  26361933 14832714000.0    562.7      5.1                      c = int(line[-1])
   278                                           
   279  26361933 35410388000.0   1343.2     12.1                      coords[:, sbin] = b
   280  26361933 12380386000.0    469.6      4.2                      data[sbin] = c
   281                                           
   282  26361933 10806356000.0    409.9      3.7                      sbin += 1
   283                                                               
   284  26361990 41117343000.0   1559.7     14.1                  progress_bar.update(1)
```

However, note that the progress bar takes about 1/3 of the time.

GallegoSav:
Thanks for the profiling study @israelmcmc ! Indeed it was expected that this part is long but I was not expected 1/3 of the time for the progress bar. I added it just to see the reading progression but if it takes that much time I'll maybe replaced it by a classic iteration incrementation. 
Concerning the RAM however the critical part is in the lines ` dr = Histogram(axes, contents=data)` and `dr_area = dr * dr.expand_dims(counts2area, 'Ei')`

PR #70, opened 2023-07-10
Improvement of the memory usage in the image deconvolution

hiyoneda:
To save memory usage, the calculation steps are revised.
- Coordinate Conversion matrix is introduced
- The model map is rotated each iteration. Now the response should be loaded only once.
- The expected counts are calculated at the initialization and in the post-process. In this way, the E-step can be skipped.
- I implemented an option to determine whether the response is loaded on the memory (or loaded at each calculation).
- Added a notebook explaining how to analyze the 2-s GRB data in miniDC2.


israelmcmc:
@hirokiyoneda Thanks for working on this. I run your notebook with _almost_ no issues, I just had to add this:
```
for label in grb_data.binned_data.axes[1:].labels:
    
    grb_data.binned_data.axes[label] = response.axes[label]
    bkg_data.binned_data.axes[label] = response.axes[label]
```

I'm not sure exactly where the issue is, but I think this is because I'm using a slightly newer version of histpy that makes a better job when comparing axes to check for the units as well. I don't think it has anything to do with this PR, so it can be merged, I'll just open an issue about updating the data IO module (heads up @ckarwin). 

Besides that, two small comments:
- Can you move your notebook from test_data to tutorials, please?
- At the beginning of the notebook, here `This notebook focuses on the image analysis with the latest image deconvolution code (on 2023-07-10)`. Can you please quote the hash of the commit you are referring instead of the date?

hirokiyoneda:
I moved the notebook from test_data to tutorials, and added the commit hash. 
Actually, I got the same error about the axes comparison. In the modified notebook, I used "_modify_axes" to avoid it, while this functionality should be removed once it is solved.

israelmcmc (approved):
Thanks @hiyoneda. Looks good to me, I'll merge

PR #71, opened 2023-08-01
Add 3ML Band function with same parameters as megalib

eneights:
Added custom Band function (Band_Eflux) that takes same parameters as those in megalib source files


israelmcmc:
Thanks @eneights . It looks good to me, except that I think it should be "particle flux" or just "flux" instead of "energy flux". If you agree, it's jut a matter of fixing the documentation. 

I run this quick test:
```python
from cosipy import Band_Eflux
from scipy.integrate import quad

a = 1e2
b = 1e4
K = 1

spec = Band_Eflux(alpha=-1,
                  beta=-2,
                  E0=500,
                  K=K,
                  a=a,
                  b=b)

print(quad(lambda e: spec.evaluate_at(e) * e, a, b)) # Energy flux

print(quad(lambda e: spec.evaluate_at(e), a, b)) # Flux
```

Which results in 
```
(573.9853040642165, 2.498720039464673e-06)
(1.0, 3.697667766561933e-09)
```

Beside that, I think this should be moved to 3ML. Also, at some point we should clean up the spectral fitting NB, they are getting hard to follows because they have multiple version of the same code within the same notebook. We can merge this now though and I'll open issues to remember us to do this. I'll just wait for the fix in the function docstring.

cosipy/threeml/custom_functions.py
```diff
@@ -0,0 +1,91 @@
...
+from past.utils import old_div
+from scipy.special import gammainc, expi
+from scipy import integrate
+import numpy as np
+import math
+
+class Band_Eflux(Function1D, metaclass=FunctionMeta):
+    r"""
+    description :
+        Band model from Band et al., 1993 where the normalization is the energy flux defined between a and b
```
israelmcmc, line 14:
```suggestion
        Band model from Band et al., 1993 where the normalization is the flux defined between a and b
```

cosipy/threeml/custom_functions.py
```diff
@@ -0,0 +1,91 @@
...
+    latex : $ A \begin{cases} x^{\alpha} \exp{(-\frac{x}{E0})} & x \leq (\alpha-\beta) E0 \\ x^{\beta} \exp (\beta-\alpha)\left[(\alpha-\beta) E0\right]^{\alpha-\beta} & x>(\alpha-\beta) E0 \end{cases} $
+    parameters :
+        K :
+            desc : Normalization (energy flux between a and b)
```
israelmcmc, line 18:
```suggestion
            desc : Normalization (flux between a and b)
```

israelmcmc (changes requested):
@eneights Based on what I understood from the last meeting, these are the only changes that these MR needs. If you are OK with them, just approve them and this can be merged.

israelmcmc (approved):
Thanks @eneights, I'll merge this.

PR #78, opened 2023-09-21
Clean up SpectralFit.ipynb

eneights:
Update spectral fit notebook to be compatible with spacecraftfile.
Also fix small mistake in SpacecraftFile.ipynb


cosipy/response/PointSourceResponse.py
```diff
@@ -74,19 +74,27 @@ def get_expectation(self, spectrum):
             spectrum_unit = spectrum.k.unit
         elif isinstance(spectrum, Line) or isinstance(spectrum, Quadratic) or isinstance(spectrum, Cubic) or isinstance(spectrum, Quartic):
             spectrum_unit = spectrum.a.unit
-        elif isinstance(spectrum, Powerlaw_Eflux) or isinstance(spectrum, Log_normal):
+        elif isinstance(spectrum, Powerlaw_flux) or isinstance(spectrum, Powerlaw_Eflux) or isinstance(spectrum, Log_normal) or isinstance(spectrum, Band_Calderone)\
```
israelmcmc, line 77: It seems that you are reverting @lommler's changes here, or is it something else?
https://github.com/cositools/cosipy/pull/28/files

Was that on purpose? If so, did you find an issue with the previous version? Was it needed to clean up the SpectralFitting exampple?

docs/tutorials/SpectralFit.ipynb
```diff
@@ -3,73 +3,382 @@
...
+    "# Spectral fitting example (GRB)"
    ]
   },
   {
    "cell_type": "markdown",
    "id": "e7df3443-3ce1-43f3-90b5-1bceb7bc9af0",
    "metadata": {},
    "source": [
-    "**To run this, you need FlatContinuumIsotropic.LowRes.binnedimaging.imagingresponse.area.nside8.cosipy.h5.zip in cosipy/cosipy/test_data. To fit the megalib simulation, you need grb_bkg_binned_data.hdf5, grb_binned_data.hdf5, & bkg_binned_data.hdf5 in cosipy/docs/tutorials/data, and to fit the source injector data, you need GRBdata.h5, GRBsignal.h5, and GRBbkg.h5 in cosipy/docs/tutorials/data.**"
+    "**To run this, you need the detector response (FlatContinuumIsotropic.LowRes.binnedimaging.imagingresponse.area.nside8.cosipy.h5), binned data (grb_bkg_binned_data.hdf5, grb_binned_data.hdf5, & bkg_binned_data_full.hdf5), and orientation file (20280301_first_2hrs.ori) saved in cosipy/cosipy/test_data. These files can be found on the COSI Pipeline Google Drive.**\n",
```
israelmcmc, line 18:
```suggestion
    "**To run this, you need the detector response (FlatContinuumIsotropic.LowRes.binnedimaging.imagingresponse.area.nside8.cosipy.h5), binned data (grb_bkg_binned_data.hdf5, grb_binned_data.hdf5, & bkg_binned_data_full.hdf5), and orientation file (20280301_first_2hrs.ori) saved in cosipy/cosipy/test_data. These files can be found on the COSI Pipeline Google Drive (Data/Data Challenges/Mini-DC).**\n",
```

docs/tutorials/SpectralFit.ipynb
```diff
@@ -13683,99 +3382,251 @@
...
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "fig,ax = plt.subplots()\n",
     "\n",
-    "for i, e in enumerate(energy):\n",
-    "    flux_inj[i] = K_inj * band_megalib(e, alpha_inj, beta_inj, xp_inj)\n",
+    "ax.plot(energy, energy*energy*flux_median, label = \"Best fit\")\n",
+    "ax.fill_between(energy, energy*energy*flux_lo, energy*energy*flux_hi, alpha = .5, label = \"Best fit (errors)\")\n",
```
israelmcmc, line 3497: Comparing the fitted results with the data counts is a good idea. However, instead of scaling  the injected counts by the an approximate exposure, you can instead compare the counts directly with the expected counts as obtained from the best-fit spectrum. This would have to go into a second plot, since the y axis will be the integrated counts (not flux). 

The advantages of doing it this way are:
- You can quantitatively obtain a goodness of fit, for example, using a chi2 test.
- The expected counts would be exact, since you'll be using all the machinery in cosipy that convolves the spectrum with the detector response. Your approximation is sound, and very good for this example, but it doesn't work in general when the orientation is changing --like for the Crab.

docs/tutorials/SpectralFit.ipynb
```diff
@@ -13683,99 +3382,251 @@
...
-    "               for par in results.optimized_model[\"source\"].parameters.values()\n",
-    "               if par.free}\n",
+    "              for par in results.optimized_model[\"source\"].parameters.values()\n",
+    "              if par.free}\n",
     "\n",
     "if 'l' in parameters:\n",
     "    del parameters['l']\n",
     "if 'b' in parameters:\n",
     "    del parameters['b']\n",
+    "if 'ra' in parameters:\n",
```
israelmcmc, line 3400: Can you remind me why you need this extra linea of code? I think we talked about it already, but can't remember. It looks hard to maintain.

israelmcmc (changes requested):
Thanks @eneights! The notebook runs for me, and it looks much better! I added some comments and suggestion before merging, please let me know what you think or if you have questions.

cosipy/response/PointSourceResponse.py
israelmcmc: Ok, thanks. I opened issue #82 to address later.

eneights: Fixed this!

docs/tutorials/SpectralFit.ipynb
israelmcmc: In general, if the uncertainty on the location of the source is comparable or greater than the angular resolution, setting the coordinates free and including them during the error propagation is the correct thing to do. It seems to me that this is rather a bug. 

For this PR, I rather leave the code using all free parameters during the propagation by default, since that the correct thing to do. Instead, can you please open an issue elaborating on the error that you get? That'll have to be solved later.

That said, a somewhat better way to obtain only the parameters of the spectrum is:

```
parameters = {par.name:results.get_variates(par.path)
              for par in results.optimized_model.source.spectrum.main.shape.parameters.values()
              if par.free}
```

This picks up the parameters specifically to the spectrum, independently of how the sky location is parametrized.

israelmcmc (approved):
Thanks @eneights , I'll merge it.

PR #92, opened 2023-10-19
Fix PointSourceResponse.py

eneights:


israelmcmc:
Small change, looks good to me. Merging. Thanks @eneights

Issue #104, opened 2023-11-16
Typo in SpectralFit example notebook 


ckarwin:
@eneights The file FlatContinuumIsotropic.LowRes.binnedimaging.imagingresponse.area.nside8.cosipy.h5 needs '.gz' at the end. Note that this needs to be corrected in two places in the notebook.

ckarwin:
This issue has been addressed with PR#105.

PR #105, opened 2023-11-16
Update response filename in SpectralFit.py

eneights:
Changing the response filename to be downloaded from wasabi to address issue #104. 
The file on wasabi is zipped, so the user needs to unzip the file manually before running the rest of the notebook


ckarwin (approved):
Thanks, Eliza. I'm also going to open an issue to see if the response class can take h5.gz or h5.zip files.

PR #108, opened 2023-11-17
Make the creation of dwell time map quiet and  clean; fix filename issue in the FullDetectorResponse module; add fast ts fit module

Yong2Sheng:
# Update 1 -- `SpacecraftFile` and `FullDetectorResponse`
According to  #88, I added two parameters (`quiet` and `save`) for `get_target_in_sc_frame` and `get_dwell_map` methods:

1. `quiet` controls if intermediate messages during calculation are printed. Default is `False`.
2. `save` controls if the target path in the SC frame and the dwell time map are saved. Default is `False`.

The issue caused by `FullDetectorResponse.open` method when the input is a `pathlib.Path` object instead of `str` is fixed. Now the `filename` will be converted into a `pathlib.Path` object regardless of the type of input.

# Update 2 -- `FastTSMap`

I also added two new modules: `fast_norm_fit.py` and `fast_ts_fit.py`.
1. `fast_norm_fit.py` is taken from the BusrtCube tool (simply copy and paste), so it doesn't need review (I guess so?). @israelmcmc please let me know what we should do to credit this module.

2. `fast_ts_fit.py` is the one I wrote to compute the parallel ts computation based on `fast_norm_fit.py`, which requires review.

3. The tutorial notebook is located at `/docs/tutorials/Parallel_TS_map_computation.ipynb`. It contains the necessary cells with simple explanatory comments to run through the whole process. I will update it in future PR with more text in detail to explain what it's actually doing. Let me know if you have any questions regarding this.

4. Now `FastTSMap` uses all the available cores to run the parallel computation. Please let me know if it significantly slows down your machine.

5. The data are upload to the folder `/cosipy/test_data`:
  - new_healpix_rsp_Binned_Bkg_2s_model.hdf5
  - new_healpix_rsp_Binned_Cosmic2s.hdf5
  - new_healpix_rsp_Binned_protoGRB.hdf5

 I already set up the path to the files in `Parallel_TS_map_computation.ipynb`, but if it doesn't work on your machine, please update the path accordingly.

Thanks!


ckierans:
Hi Yong, I've looked through this pull request. Mostly I focused on the Parallel_TS_map_computation jupyter-notebook, and I have a few comments but these are mostly cosmetic vs functional.

- It will be important to test this with the SMEX model once we have a GRB sim to do this
- Is there a way to get the orientation information from the ori file instead of hardcoding it? For example, if you know the time of the burst, can you not just take the range of pointings during that time?
- When running the ts_results = ts.parallel_ts_fit(... results cell, I am getting 16 copies of lots of warnings. These mostly seem related to astromodels, and 3ML
- Can you include the option to select the number of cores to run with in the parallel_ts_fit, since using all of my cpu cores is limiting if I'm trying to do other work in parallel?
- For the final image when you plot the known location with longitude and latitude, can you instead plot an x at the correct location so it's clear which pixel this corresponds to since the arrow isn't as obvious (is the point of the arrow the actual true location?)
- Just FYI, it took my computer with 16 GB of memory and 16 cores 21 minutes to run and produce an image

ckierans (approved):
Approving for review since the functionality of the parallelized TS map works. Added a few cosmetic suggestions in my comment above, and will open issues for the most important ones.

PR #116, opened 2023-11-29
Crab spectral fitting in galactic coordinates

eneights:
I added a notebook to fit the spectrum of the Crab in galactic coordinates, and added the scatt map to COSILike.py


ckarwin:
Hi Eliza, I check your PR and everything runs fine. Here are some small changes before merging:

Please specify which BG you’re using in the notebook. 

Fix comments below (still refers to GRB): 

“Read in the spacecraft orientation file & select the beginning and end times of the GRB”

“Create BinnedData objects for the GRB only, GRB+background, and background only. The GRB only simulation is not used for the spectral fit, but can be used to compare the fitted spectrum to the source simulation”

In three3ml: “coordsys=None” should be described in the doc string

I still need to check a couple more things.

PR #121, opened 2023-12-18
Fix line in spectral fit notebook

eneights:
Added line to specify coordinate system of binned files to be compatible with current version of COSILike.py


ckarwin (approved):
I checked this PR and everything is working ok. 

To add some more context, the notebook for the spectral line fit (based on the Compton sphere) uses binned data that was created before the last updates to the dataIO class. Consequently, the binned data does not have coordinate system information. This leads to an error when defining COSILike if one does not pass the coordinate system by hand.  With this PR, the coordinate system is now specified when defining COSILike.

PR #122, opened 2024-01-05
Add GRB spectral fit notebook with SMEX mass model

eneights:


ckarwin:
Hi @eneights I checked your PR and everything looks good. Nice work! I have just one small issue that I ran into:
The response file on wasabi has a '.zip' extension, which should be included at the top of the notebook when you give the needed files. In this case you might also want to specify that the response file needs to be unzipped. Other than that, I think the PR is ready to be merged.

ckarwin (approved):
I checked all the changes and everything looks good.

PR #123, opened 2024-01-07
Realistic 511 keV model spectral fitting

fieldrog:
adding realistic 511 model case to diffuse511 spectral fitting notebook


fieldrog:
PR into Chris' PR instead

PR #125, opened 2024-01-18
correct FullDetectorResponse.py for the reading of sparse parameter

GallegoSav:
Just a small correction as in order to read correctly the sparse parameter in the response file. I also added a option in case the SP parameter is empty in the file.

### DIFF
--- a/cosipy/response/FullDetectorResponse.py
+++ b/cosipy/response/FullDetectorResponse.py
@@ -190,21 +190,24 @@ class FullDetectorResponse(HealpixBase):
                 elif key == 'SA':
                     area_sim = float(line[1])
                     
-                elif key == "SP" and line[1]!="true" and line[1]!="false" :
-                    norm = str(line[1])
+                elif key == "SP" :
                     
+                    try :
+                        norm = str(line[1])
+                    except :
+                        print(f"norm not found in the file ! We assume {norm}")
+
                     if norm =="Linear" :
                         emin = int(line[2])
                         emax = int(line[3])
                         
-                elif key == "SP" and line[1]=="true" :
-                    sparse = True
-                    
-                elif key == "SP" and line[1]=="false" :
-                    sparse = False
+             
 		
                 elif key == "MS":
-                    sparse = bool(line[1])	
+                    if line[1] == "true" :
+                        sparse = True
+                    if line[1] == "false" :
+                        sparse = False
 
                 elif key == 'AD':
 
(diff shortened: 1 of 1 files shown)


cosipy/response/FullDetectorResponse.py
```diff
@@ -190,21 +190,24 @@ def _open_rsp(cls, filename, Spectrumfile=None,norm="Linear" ,single_pixel = Fal
...
-                elif key == "SP" and line[1]=="true" :
-                    sparse = True
-                    
-                elif key == "SP" and line[1]=="false" :
-                    sparse = False
+             
 		
                 elif key == "MS":
-                    sparse = bool(line[1])	
+                    if line[1] == "true" :
```
israelmcmc, line 207: Was `bool('true')`/ `bool('false')`  failing? I'd had thought it should work. Or what was the issue?

cosipy/response/FullDetectorResponse.py
```diff
@@ -190,21 +190,24 @@ def _open_rsp(cls, filename, Spectrumfile=None,norm="Linear" ,single_pixel = Fal
                 elif key == 'SA':
                     area_sim = float(line[1])
                     
-                elif key == "SP" and line[1]!="true" and line[1]!="false" :
-                    norm = str(line[1])
+                elif key == "SP" :
                     
+                    try :
```
israelmcmc, line 195: If I understand this correctly, this is mostly so it is backwards compatible? Why do some files have an empty SP and some don't?

israelmcmc (commented):
Thanks @GallegoSav . I think I can merge this quickly, I just wrote a couple questions to make sure I understood the code.

cosipy/response/FullDetectorResponse.py
GallegoSav: Yeah me too but it is actually always return `True` , even bool("false")

cosipy/response/FullDetectorResponse.py
GallegoSav: I think Andreas said he used a new(old) version or maybe this is a bug. Anyway this line is more to check if it was empty or not and warn the user

cosipy/response/FullDetectorResponse.py
israelmcmc: Right! I thought I had checked, but saw it wrong. Never mind, thanks.

cosipy/response/FullDetectorResponse.py
israelmcmc: Ok, thanks.

israelmcmc:
This looks good, I'll merge it. Thanks, @GallegoSav

israelmcmc (approved):

PR #126, opened 2024-01-23
Sphinx documentation

israelmcmc:
Automatic sphinx documentation. 

 Currently, it is built automatically from the `docs` branch, and deployed here: https://cositools.github.io/cosipy/
Eventually, we can use the releases on the main branch and deployed to somewhere like https://cosi.ssl.berkeley.edu/cositools/cosipy

Here's how to compile it locally: https://cositools.github.io/cosipy/install.html#compiling-the-docs

Currently, you'll need to locally merge this branch (`docs`) into your branch before compiling in order to test changes to your docstrings and notebooks.

Note: The `#` and `##` in the markdown cells of your notebook now do matter. These correspond to the title and subtitles shown in the table of contents. There should be only one `#` for the title. I fixed a couple of notebook in order to make it work. I'm also overriding the title of the notebooks for now in the table of contents, but eventually they need to be set by the notebook itself.

As for the API section, I'm currently using autodocs to document the each module automatically. You can change the behaviour by going to the .rst of the module, for example: https://github.com/cositools/cosipy/blob/docs/docs/api/response.rst?plain=1

I'll leave this as draft for now because I need to add more explanatory text.


ckarwin (approved):
Ready to be merged

Issue #134, opened 2024-02-15
COSIpy Installation - missing files (in setuptools include_package_data)


avalluvan:
Hi team,
I have installed COSIpy within a virtual python environment (v3.10) over WSL (Ubuntu 22.04). When I tried to run the Data I/O notebook out-of-the-box, the following input "housekeeping" files were missing:
![Screenshot 2024-02-13 215645](https://github.com/cositools/cosipy/assets/62253557/5404d487-97a6-4aaa-b2b5-db028792aadd)
![Screenshot 2024-02-13 215745](https://github.com/cositools/cosipy/assets/62253557/2c45f9c4-b37b-4a07-9f0d-02de63f9e845)
![Screenshot 2024-02-13 215813](https://github.com/cositools/cosipy/assets/62253557/4fddbe6f-5e0d-49b0-a772-06b3a773c3b3)

Performing a simple `wget` from [the source ](https://github.com/cositools/cosipy/tree/main/cosipy/data_io/housekeeping_files) fixed the issue.

ckarwin:
Thanks for catching this, @avalluvan. The missing housekeeping files are actually not currently needed, and so I've removed this from the code. I just submitted a PR to resolve this issue.

StevenBoggs:
I am having the same issue in MacOS, M1 Max. How exactly do we implement  @ckarwin 's  fix to get past this error in the testing? (Or do we wait for the public release?)

StevenBoggs:
I think I figured it out, I made @ckarwin 's modifications in my local DataIO.py file. :-)

ckarwin:
Yes, thanks @StevenBoggs. That's the fix for now until we make an updated release.

israelmcmc:
@ckarwin I made a new alpha release (v0.0.2a3) to include your changes and fix this particular problem with the DataIO. 

I still need to see which other files to list in [include_package_data](https://setuptools.pypa.io/en/latest/userguide/datafiles.html), but this should work for now.

ckarwin:
Yes, sounds good.

israelmcmc:
I'm closing this because we are now not using the test_data for the examples at all, only data from wasabi.

PR #135, opened 2024-02-15
removed unused housekeeping directory from data_io

ckarwin:
The housekeeping directory in data_io is not needed at this time. I kept it there from when I was working on the parameter configurator. It has now been removed. This should resolve issue #134.


israelmcmc:
Thanks @ckarwin . I'm merging this but I'll keep the issue open. See my [comment](https://github.com/cositools/cosipy/issues/134#issuecomment-1946259349) there.

Issue #139, opened 2024-02-20
Root call in Crab Imaging notebook


StevenBoggs:
Got through all 3 Crab imaging workbooks fine until near the very end of Crab-DC2-ScAtt-Upsampling:

<img width="1207" alt="Screenshot 2024-02-19 at 7 07 49 PM" src="https://github.com/cositools/cosipy/assets/160159429/ecbeaea3-be6d-4dcf-b318-eac868ffce25">

First I've seen this error, is Root expected for running these notebooks in DC2?

ckarwin:
Hi @hirokiyoneda, I think this is resolved now. Can we close this issue?

PR #143, opened 2024-02-22
Updated image deconvolution NBs 

hiyoneda:
- I updated image deconvolution nb with wasabi util function
- I removed ROOT dependency from Crab-DC2-ScAtt-Upsampling.ipynb 
  - ROOT was used for the interpolation of a graph.
  - Then, `scipy.interpolate` is used instead of it.
  - #139 is now resolved.


ckarwin (approved):
I checked the changes and all looks good.

PR #144, opened 2024-02-23
Add cells to download files in grb & crab spectral fit notebooks

eneights:
Add cells in the grb & crab spectral fitting notebooks to download the necessary files from wasabi

### DIFF
--- a/docs/tutorials/spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb
+++ b/docs/tutorials/spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb
@@ -7,7 +7,7 @@
    "source": [
     "# Diffuse 511 Spectral Fit in Galactic Coordinates\n",
     "\n",
-    "This notebook fits the spectrum for the 511 keV emission in the Galaxy. It can be used as a general template for fitting diffuse/extended sources in Galactic coordinates. For a general introduction into spectral fitting with cosipy, see the continuum_fit tutorials.<br>\n",
+    "This notebook fits the spectrum for the 511 keV emission in the Galaxy. It can be used as a general template for fitting diffuse/extended sources in Galactic coordinates. For a general introduction into spectral fitting with cosipy, see the continuum_fit tutorial.<br>\n",
     "\n",
     "This notebook uses two 511 keV emission models, first a test model and then a realistic multi-component model. \n",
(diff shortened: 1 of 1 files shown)


ckarwin:
Hi @eneights, Thanks for making these changes. I reviewed your PR, and all the new cells are working. There are still some small details that I think need to be updated though, specified below. Also, it looks like the line fit example has not been updated yet. Were you planning to do this as well? 

Both Crab and GRB notebooks:

1)  > To run this, you need the following files, which can be found on wasabi and downloaded using the first few cells of this notebook:

Most users won't have access to wasabi, and so I think you can just say that the files need to be downloaded using the first few cells of the notebook. 

2) You still have your personal data path in the first cell of the crab notebook. I think the default can just be ".", which is the working directory. Or "path/to/files", as you have in the GRB notebook. 

3) Can you please also add the file sizes for each cell, in order to give users a sense of how long the download will take? They can be found on wasabi. 

4) I think it would be sufficient to add a comment before the cells specifying that the files should only be downloaded if not previously downloaded, instead of repeating this for each cell.

eneights:
Thanks @ckarwin! I made those changes. 

The point source line fitting notebook should probably be deleted, do you want me to do that? It's for the Compton sphere and the extended source 511 notebook replaces it.

ckarwin:
Thanks for making the changes, @eneights. 

Ok, I see. I think it's probably ok to remove the line fitting notebook. Let's double check with @israelmcmc, do you see any problems with this?

israelmcmc:
I also agree it's good to remove it. Conceptually, the jump from a PL or band function to a Dirac delta (or a narrow Gaussian) is not that big, so I think the users will not get lost if the line emission analysis is introduced together with the extended source emission ---which is the realistic use case anyways. The 511-emitting "GRB" had its purpose when were trying to figure out the implementation and it allowed us to tackle one problem at a time, but at this point, it's better to reduce the number of notebooks that we need to maintain.

ckarwin:
Ok, @eneights, please go ahead and remove the line fit NB, and then I can merge this PR. 

In the second line of the extended source NB it says:

> For a general introduction into spectral fitting with cosipy, see the continuum_fit and line_fit tutorials.

Can you also delete:

> and line_fit tutorials

ckarwin (approved):
Thanks, @eneights.

PR #146, opened 2024-02-26
deleted old GRB imaging notebooks

hiyoneda:
I deleted old GRB notebooks. They were used for testing in miniDC2, but the codes on them are outdated and are useless now.


ckarwin (approved):
Thanks, @hiyoneda.

PR #151, opened 2024-03-01
Fix documentation link and only print one log-likelihood warning

eneights:
Updated index.rst to fix link to GRB spectral fitting notebook.

Updated COSILike and spectral fitting notebooks to only print one warning about the log-likelihood per fit. This is to avoid having the same warning printed repeatedly every time the log-likelihood is calculated in the example notebooks and documentation.


ckarwin:
Very nice @eneights! I ran through the GRB notebook and it works well. I also applied these changes to the extended source NB, in order to prevent unwanted output (Eliza added this to her PR). Note that for the extended source NB with multiple sources, there is a bunch of additional output from 3ML pertaining to the updated log values at each iteration of the fit. In order to silence this I also used the following right before the fit cell:
```
from threeML import update_logging_level
update_logging_level("CRITICAL")
```
 
Now the spectral fit NBs no longer have tons of output from the fit, which will make the tutorials much nicer and easier to read.

ckarwin (approved):

PR #156, opened 2024-03-05
Add psutil, a new dependency, to the mocked packages

### LINKED ISSUE #155: New dependency: psutil
I recent PR added a psutil as a new depency: https://github.com/cositools/cosipy/pull/145

This broke the docs build since psutil needs to be added to the mocked packages.

I haven't checked if it broke the installation, since maybe psutil was already a secondary dependency.

israelmcmc:
Closes #155

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -66,6 +66,7 @@ autodoc_mock_imports = ["histpy",
                         'pandas',
                         'tqdm',
                         'scipy',
+                        'psutil',
                         'awscli']
 
 # There seems to be a conflict between unittest.mock (used by sphinx) and metaclasses


ckarwin:
Thanks, @israelmcmc! Yeah this fixes the issue. I ran the full cosipy install (using the psutil branch), staring with a new environment, and everything worked ok for me.

One small note: The detector response in the API section has a slightly different style than the other sections. I think this is b/c `response.rst` uses `autoclass` separately for each class, whereas all the other modules use `automodule` . It's a small difference, but do you want to modify this to make things the same style, or leave how it is?

Otherwise, I think this can be merged.

ckarwin (approved):

PR #171, opened 2024-03-24
Fast TS map and add support for regional fitting and use `scatter` intead of `text` to plot the true location.

Yong2Sheng:
These changes are made for introducing the TS map method into my localization pipeline and optimizing the plotting method.

# Update 1
In order to accelerate the ts map fitting in localization simulation, I modified `fast_ts_fit.py` to fit a specific disc region instead of the full map.

# Update 2
1. `hp.projtext` will place an offset between the text and the coordinate you give, which caused a noticeable systematic offset between the localized 90% containment region and the true location.

2. Now it uses `hp.projscatter`, and the markers are placed at the exact location. If you run [ts map tutorial notebook](https://github.com/Yong2Sheng/cosipy_SpacecraftFile/blob/fast_ts_map_galactic/docs/tutorials/ts_map/Parallel_TS_map_computation_DC2.ipynb), you will find that the containment region is more consistent with the true location of the source than before.


israelmcmc:
Thanks @Yong2Sheng . Can you please separate this into two PRs? The plotting change is simple and I'd like to include it in DC2, but I rather not merge the other one right before the release.

Yong2Sheng:
Sure! I will make a new branch that has the changes in the plot method only!

israelmcmc:
Thanks @Yong2Sheng !

Yong2Sheng:
Hi @israelmcmc, I made another PR #172 that only contains the updates to the plotting method. I will modify this one later so it will only contain the updates to the regional fitting part.

israelmcmc:
I changed the target branch to "develop". I want to avoid adding new features to the version we're releasing today. I'll explain this further tomorrow during the cosipy meeting.

GallegoSav:
Hi @Yong2Sheng , I tried your pr and everything went well for me. That's indeed fast when you can get lot of CPUs.

However, why do you use a so fine binning in time (0.5s) for `Crab_bkg_galactic_inputs.yaml`  It is making a heavy file for the background and I don't see where you use the time in  you code  because it seems for me you are using the total 3 months bck.

In addition you only used the albedo photon background component. I tried with `total_bg_3months_binned_for_continuum.hdf5`, that's only 188Mo because time binning is 3600s. The TS map seems to converge well when we use all the background.

image
image

Yong2Sheng:
Hi @GallegoSav,

Thank you for your comments!  it’s a good idea to use larger time binning to save the file size. I will make a new version of the data and upload to Wasabi after tests. (Total background and larger time binning).

israelmcmc:
@Yong2Sheng: @GallegoSav is a good suggestion, but I think it can be done in a separate PR, if you want. This one is working properly, it's only a matter of the user inputs. The only thing preventing this PR from being merged is the unit test. 50% of the new lines need to be covered (it says 100%, but I have a tolerance of 50% for now). Please let me know if you need help with the unit tests.

GallegoSav:
> @Yong2Sheng: @GallegoSav is a good suggestion, but I think it can be done in a separate PR, if you want. This one is working properly, it's only a matter of the user inputs. The only thing preventing this PR from being merged is the unit test. 50% of the new lines need to be covered (it says 100%, but I have a tolerance of 50% for now). Please let me know if you need help with the unit tests.

@israelmcmc I think I would need help for unit test because I never done that 
image

Yong2Sheng:
Hi @israelmcmc, 

Thank you for the comments. I will work on the unit test. I will let you know if I need help on this.

israelmcmc:
@GallegoSav I was asking Yong to add the tests for this PR, but you'll need to do it as well at some point hehe. Here are some instructions: https://github.com/cositools/cosipy/blob/c8eb53631f686621d93a8b56395ce66ae3d71825/docs/dev/unit_tests.rst

Yong2Sheng:
Hi @israelmcmc @GallegoSav , the unit test of `ts_map` module has been added:

1. `fast_norm_fit.py` coverage is 80%
2. `fast_ts_fit.py` coverage is 54%

Could you please take a look? Thank you!

israelmcmc (approved):
Thanks @Yong2Sheng! Everything looks good, I'm merging it.

PR #172, opened 2024-03-24
Adding saving option and using scatter to project the true location in the TS map

Yong2Sheng:
These changes are made to optimize the plotting method of `fast_ts_fit.py` module. The new features are as follows:

# Feature 1
You can save the ts plot by specifying `save_plot`, `save_dir` and `save_name`.

# Feature 2
1. `hp.projtext` will place an offset between the text and the coordinate you give, which caused a noticeable systematic offset between the localized 90% containment region and the true location.

2. Now it uses `hp.projscatter`, and the markers are placed at the exact location. If you run [ts map tutorial notebook](https://github.com/Yong2Sheng/cosipy_SpacecraftFile/blob/TS_plot_method/docs/tutorials/ts_map/Parallel_TS_map_computation_DC2.ipynb), you will find that the containment region is more consistent with the true location of the source than before.


israelmcmc:
Thanks @Yong2Sheng . I verified that this runs on the nb (at least the GRB examples). I'll merge it.

Issue #158, opened 2024-03-11
Avoid reading the orientation file multiple times


israelmcmc:
No need to read the file multiple times, all columns can be read at the same time.

https://github.com/cositools/cosipy/blob/f8a8518bf5c771ae171233966ac9c254af77753d/cosipy/spacecraftfile/SpacecraftFile.py#L100

israelmcmc:
Closed by #177

PR #177, opened 2024-05-14
Refactor file loading to improve efficiency

krishnatejavedula:
This commit refactors the code to load data from the file only once instead of multiple times. Instead of calling  three times to extract different columns, the code now loads the file once and extracts all necessary columns from the loaded data array. This change improves code efficiency by reducing redundant file reads.

Changes made:
- Load file data only once using .
- Extract necessary columns from the loaded data array.

Closes issue #158

### DIFF
--- a/cosipy/spacecraftfile/SpacecraftFile.py
+++ b/cosipy/spacecraftfile/SpacecraftFile.py
@@ -101,9 +101,11 @@ class SpacecraftFile():
             The SpacecraftFile object.
         """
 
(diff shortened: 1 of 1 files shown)


israelmcmc:
I changed the base branch, from main to develop. This is a recent change of workflow. After that, I'm seeing an issue with the commit to be merged, so I'm going to close and reopen this to see if that fixes it.

israelmcmc:
It seems that worked.

israelmcmc (approved):
I can confirm that this doesn't change the behaviors of SpacecraftFile and that it cuts the loading time by ~40%. Thanks @krishnatejavedula! I'm merging this.

Issue #48, opened 2023-05-10
Change print() to logging.info()


israelmcmc:
When thing's settle, we'll have to go through the code and find these out.

israelmcmc:
Remaining offenders: 
```
./data_io/UnBinnedData.py
./data_io/ReadTraTest.py
./data_io/BinnedData.py
./ts_map/fast_ts_fit.py
./ts_map/TSMap.py
./threeml/COSILike.py
./response/FullDetectorResponse.py
./image_deconvolution/RichardsonLucy.py
./image_deconvolution/data_loader.py
./image_deconvolution/exposure_table.py
./image_deconvolution/deconvolution_algorithm_base.py
./image_deconvolution/image_deconvolution.py
./spacecraftfile/SpacecraftFile.py
```

Tagging POCs for awareness:
- data_io: @ckarwin 
- ts_map: @Yong2Sheng 
- threeml: @eneights 
- response: Me! @israelmcmc 
- image_deconvolution: @hirokiyoneda 
- spacecraftfile: @Yong2Sheng @saurabhmittal23 

This is a very "good first issue", so try to delegate!

ckarwin:
@krishnatejavedula Do you know if this is still an issue for any of the other modules? I think maybe ts_map and image_deconvolution?

krishnatejavedula:
Hello @israelmcmc, @ckarwin. You are right. It is still in an issue for the rest of the files and I have yet to make the changes for them. I will be doing that soon and include them in my next PR.

hiyoneda:
@israelmcmc @ckarwin, do you plan to use logging.info as an output for any data analysis? I understand that we should replace 'print()' with logging.info(), warning(), etc. for debugging and logging. However, we sometimes want to show the results in the data analysis process, e.g., the number of extracted events, the significance of the detection, the total flux, etc. Since they are not the logging info, can we still use print or pprint for them?

ckarwin:
@israelmcmc should comment as well, but I'll give my input. You can also take a look at the related discussion in the comments of PR #178. My general understanding is that logging replaces all print statements. Then anything that would normally be printed to terminal should be logged with logging.info(). The user can then set the logging configuration as desired. Note that they only need to configure the root logger, and all other modules will use the same. For example, they can set it as follows, in which case all the logging info will be written to file and printed to terminal:

```
logging.basicConfig(filename='cosipy.log', level=logging.INFO)
[...]

Issue #26, opened 2023-05-04
Standarize the coding style


israelmcmc:
At the moment we haven't paid much attention to coding standards, but this has to be revisited. 

Also add coding style checks to the CI pipelines.

israelmcmc:
Data IO class should use `._` for internal member, and properties to the ones interfacing with the user

PR #178, opened 2024-05-16
Change print() to logging.info() #48

krishnatejavedula:
- Replaced print statements with logging in various files to improve logging consistency, clarity, and debugging capabilities.
- Updated logging statements in the following files:
  - ./data_io/UnBinnedData.py
  - ./data_io/ReadTraTest.py
  - ./data_io/BinnedData.py
  - ./threeml/COSILike.py
  - ./spacecraftfile/SpacecraftFile.py

### DIFF
--- a/cosipy/data_io/BinnedData.py
+++ b/cosipy/data_io/BinnedData.py
@@ -53,8 +53,8 @@ class BinnedData(UnBinnedData):
         but it does not explicitly return it.
         """
         
-        # Make print statement:
-        print("binning data...")
+        # Log message:
+        logger.info("binning data...")
       
         # Option to read in unbinned data file:
         if unbinned_data:
@@ -72,18 +72,14 @@ class BinnedData(UnBinnedData):
             num_bins = round(delta_t / self.time_bins)
             new_bin_size = delta_t / num_bins
             if self.time_bins != new_bin_size:
-                print()
-                print("Note: time bins must be equally spaced between min and max time.")
-                print("Using time bin size [s]: " + str(new_bin_size))
-                print()
+                logger.info("Note: time bins must be equally spaced between min and max time.")
+                logger.info("Using time bin size [s]: " + str(new_bin_size))
             time_bin_edges = np.linspace(min_time,max_time,num_bins+1)
 
         if type(self.time_bins).__name__ == 'list':
             # Check that bins correspond to min and max time:
             if (self.time_bins[0] > min_time) | (self.time_bins[-1] < max_time):
-                print()
-                print("ERROR: Time bins do not cover the full selected data range!")
-                print()
+                logger.error("ERROR: Time bins do not cover the full selected data range!")
                 sys.exit()
             time_bin_edges = np.array(self.time_bins)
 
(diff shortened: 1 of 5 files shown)


ckarwin:
@krishnatejavedula I went over all the changes in this PR and everything looks good. I tested the code for the dataIO using new unit tests that I've prepared locally, and everything works fine. I also tested the dataIO tutorial notebook with no problems. 

The only issue is that with the current (default) configuration, all of the logging info is effectively lost, i.e. it's not printed to terminal, nor saved to file. However, it is highly beneficial to have both. The terminal output is needed to update users on the flow of the analysis (when running from command line or in a notebook), and writing to a log file is useful for future reference. 

Only the root logger needs to be configured, since all the lower level loggers at module level eventually forward their messages to its handlers. I think a good place for the root logger is in `cosipy/config/configurator.py`. So in this file I would suggest that you add the following lines:
```
import sys
logging.basicConfig(filename='cosipy.log', level=logging.INFO)
logging.getLogger().addHandler(logging.StreamHandler(sys.stdout))
```
This will print all logging info to stdout, and save it to a log file called cosipy.log (in the working directory). Note that by default, the logging level will be INFO. In the future we can set this as an input, e.g. to allow users to set it to `DEBUG` if needed, or a lower verbosity (`ERROR`) if they don't want any logging info for some reason. 

I'll get @israelmcmc in the loop here to see what he thinks, b/c this ultimately relates to issue #26, and having a standard logging across all modules. 

After making the above changes, I think this PR can be merged.

ckarwin:
Note that `DataIO.py` already reads the input yaml file using `Configurator()`, and so it automatically uses its logger config.

ckarwin:
Actually will need to think about how to setup the root logger somewhere.

israelmcmc:
Thanks for reviewing @ckarwin!

TL;DR: I think this one can be merged now.

It's best practice that the root logger is not configured by the library but by the user (discussed [here](https://docs.python-guide.org/writing/logging/#logging-in-a-library)). Cosipy can be used as part of an even larger codebase, and who knows what they want to do then. This flexibility and customizability by the final user is actually the main reason to use logging instead of print statements. 

By default, only warnings are printed to stderr (discussed [here](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)). The rationale, I assume, is that good code works like a black box, and you only want to be bothered if something goes wrong and needs attention. Also, the default is stderr and not a file because that way the user can redirect all the logs to a single place without having to worry about what the internal pieces of code are doing do.

That said, I agree that using the INFO level is beneficial for most of our purposes, especially at this point in the development. We can simply add the following to the notebooks and our executable scripts (applications):
```
import logging
logging.basicConfig(level=logging.INFO)
```

Our applications should have a `-v` option that controls the verbose, which will be translated to e.g. 0 = warning only, 1 = info, 2 = debug.

I rather not output to a file by default, because then:
1. You won't see the result in a jupyter notebook cell, which is awkward when working with interactive notebooks
2. For an executable scripts, you lose the ability to redirect stderr to a file of your choice with `2>` in the command line.

It would be convenient for our application to have an optional flag to save the logs to a file though.

This should all be part of the global config yaml file as well.

And yeah, I should write all of this down in a style guide so it can be discussed and followed. 

In any case, I think this is all for future PRs, I think this one can be merged now.

ckarwin (approved):
This PR has been checked and discussed. Good to merge.

israelmcmc:
Yeah, maybe outputting to both a file and stderr is a good idea. I guess we care more about reproducibility than the regular software, and this might be a saver.

krishnatejavedula:
Thank you @israelmcmc and @ckarwin for your feedback. This helps a lot. I am glad the PR is merged. I will be working on the rest of files mentioned in #48 and open a PR for them shortly so the issue could be closed.

PR #182, opened 2024-05-31
Earth occultation in the image deconvolution (ScAtt binning)

hiyoneda:
I added a functionality to consider the earth occultation in the coordinate conversion matrix. 
The earth occultation can be included in the analysis of the image deconvolution with the scatt binning + local coordinate.

**I want to emphasize that it is limited to the case that the satellite always points to the zenith. So, it is valid for the DC2 but not for general situations.**

Also, I changed the codes additionally as follows. All of them are minor changes.
- mask a part of the sky if such a region has zero exposure time.
- add a parameter `mininum_flux` in deconvolution_algorithm_base in order not to make pixels with minus flux. The default value is 0.
- modified the implementation of the function that calculates the acceleration parameter in the RL algorithm.
- use hp.smoothing for the Gaussian image smoothing instead of preparing a large matrix.

I compared the reconstructed spectrum of 3-month Crab data with/without considering the earth occultation effect.
The reconstructed flux becomes more consistent with the injected model, especially in the higher energy band where gamma rays can pass through the bottom BGO shield more easily. For example, the difference changes from -26% to 0.3% at the highest energy band.

Screenshot 2024-05-31 at 17 50 50
Screenshot 2024-05-31 at 17 50 57


israelmcmc:
Hi @hirokiyoneda. Thanks for working on this. I'm OK with all the changes you listed, but I'm on the fence about adding the Earth occultation code itself. 

For this workaround, specific to DC2, we don't need to change the code at all, just change the detector response file. Since this new code is not meant to be definitive, and the same can be achieved with a simpler solution --that doesn't involve people getting a new release-- I thought it was cleaner to leave the code as is and save us from more code that needs to be tested and maintained. What do you think?

hiyoneda:
That's a good point. I wanted to merge it into the development branch because:
- The change in the coordinate conversion matrix does not harm the DC2. While I added a new parameter, `earth_horizon_angle`, the old notebooks published for DC2 should work since `earth_horizon_angle` is not a mandatory parameter.
- To consider the earth occultation effect in a general way, I understood that we need information about the satellite altitude and position in the orientation file. So, I think that we also need to modify the orientation. I think that this code can be a good example to understand and discuss it.

But I agree that it is cleaner to keep it as it is.

One concern is that I made minor changes following the earth occultation function. Actually, my current other works depend on them. I am happy if I can merge them into the develop branch. Should I do PR including these minor changes but without the earth occultation functions? Is it a good way to remove the commit regarding with the earth occultation, probably using `git revert`?

israelmcmc:
Depending on your situation, you can use either `git rebase --onto` (see [this](https://git-scm.com/book/en/v2/Git-Branching-Rebasing#:~:text=More%20Interesting%20Rebases)) or `git cherry-pick` ([here](https://www.git-scm.com/docs/git-cherry-pick)). I'll follow up on this over slack.

You can create a branch called `e.g. earth_occ_DC2` from the commit that has the Earth occultation code, but before you made these other changes, and maybe open a draft PR with it. I agree we should keep it around as an example and discuss it when we work out the actual code that works for the general, but no need to merge it now.

hiyoneda:
Thanks a lot. I am closing this PR.  I have made a draft PR by removing unrelated changes as #184.

PR #184, opened 2024-06-03
Earth occultation in the image deconvolution for DC2 + ScAtt binning

hiyoneda:
I added a functionality to consider the earth occultation in the coordinate conversion matrix.
The earth occultation can be included in the analysis of the image deconvolution with the scatt binning + local coordinate.

**I want to emphasize that it is limited to the case that the satellite always points to the zenith. So, it is valid for the DC2 but not for general situations. Thus, this is just an example to think about how to include the earth occultation in the analysis.**

I compared the reconstructed spectrum of 3-month Crab data with/without considering the earth occultation effect.
The reconstructed flux becomes more consistent with the injected model, especially in the higher energy band where gamma rays can pass through the bottom BGO shield more easily. For example, the difference changes from -26% to 0.3% at the highest energy band.

image
image


israelmcmc:
I'm closing this in favor of #258. Thanks @hiyoneda! Please feel free to reopen it if needed.

Issue #101, opened 2023-11-14
bug in meta data when writing fits files with UnBinnedData class


ckarwin:
When writing the un-binned data to a fits file, the meta data includes the keyword "data_file". There are two problems with this.
1. If the name of the file is too long, this will cause an error (as found by @eneights). This can happen because the code is using the data file name in the inputs yaml file, which often contains a long path. 
2. This meta data doesn't apply when combining multiple fits files.

israelmcmc:
@ckarwin Can this be closed?

ckarwin:
@israelmcmc Not yet. It will be an easy fix. I'll get to it soon.

Issue #95, opened 2023-11-09
Add option to read tra file with no associated pointing information


ckarwin:
short description:
Need to add option in dataIO to read tra file with no pointing information. 

long description:
The read_tra() method requires some associated pointing information (either directly in the tra file or via an ori file). This is needed to convert the data from local coordinates to Galactic. However, sometimes simulations are ran just in local coordinates, with no pointing information. In this case, the data cannot be converted to Galactic coordinates. An option needs to be added to allow for this.

israelmcmc:
@ckarwin What's the status on this one? Or maybe it's not needed anymore?

ckarwin:
@israelmcmc I think it's still needed. I'll get this implemented soon.

israelmcmc:
Thanks. Yeah, that's alright, just trying to keep the list of issues up to date

PR #186, opened 2024-06-04
Unit Tests for DataIO Class

ckarwin:
I added unit tests for the dataIO class, so that every class now has >=95% coverage. Specifically, the coverage increase is as follows:

BinnedData.py: 13% -> 95%
ReadTraTest.py: 15% -> 95% 
UnBinnedData.py: 62% -> 95%
make_plots.py: 14% -> moved to BinnedData class, since it's the only class using it. 

Note that the coverage for ReadTraTest.py will only be for users with MEGAlib installed locally. If MEGAlib is not installed, there will be no problem -- it will just skip the test. The test file directory already has a file generated with ReadTraTest. The class is kept here now for convenience, for generating the test file. It can probably be removed in the future, but at least for development it is important to always cross check the cosipy tra parser with the results from MEGAlib, and so it's probably best to keep for now.

For the UnBinnedData class I updated the test files, but they are very similar. I'm using the first 10 seconds of the Crab sim from mini-DC2. Previousy it was the first 1 second or so. The test files are still very light, but it's more helpful to have a bit more time, specifically for testing that the ori file interpolation (in UnBinnedData.py) agrees well with the actual values from the tra file, which it does. 

This PR also resolves the following issues:
issue #101 
issue #95


ckarwin:
As seen in the automated coverage report, the coverage for ReadTraTest.py is 19.75% without MEGAlib. Agian, it's 95% with MEGAlib locally installed, and this file is mainly for development purposes.

israelmcmc:
Yay! The automatic tests and coverage worked! And it shows the increase in coverage. I think this is the first real test with a PR besides mine. 

It looks good to me. And since this is for the additions of tests (thanks, @ckarwin!), which were run automatically, there's not much to test.

One comment though: While `show_plots=True` is an intuitive default value for the function that have a `plot` in the name, I think it should be `False` by default for all others (`get_binned_data`, `get_raw_spectrum`, `get_raw_lightcurve`).

ckarwin:
Thanks for the quick review @israelmcmc. Yeah, the automatic tests and coverage report are very nice to have! 

Ok, that's reasonable. I changed the defaults as suggested.

israelmcmc (approved):
Thanks for the changes. Looks good to me. We can merge it as soon as the new automatic unit tests finish.

PR #188, opened 2024-06-11
Major updates on the image deconvolution modules

hiyoneda:
I have updated the image deconvolution modules. It changes the overall structure of the image deconvolution, which is a major update.

The basic idea is to separate deconvolution algorithms from the actual data format and the definition of a model to be reconstructed. 
Now, the deconvolution algorithm, data, and model have their corresponding base classes (DeconvolutionAlgorithmBase, DataLoaderBase, and ModelBase, respectively), and actual implementation should be done by inheriting a corresponding base class.
Then, the ImageDeconvoltuion class uses these subclasses and performs image deconvolution.

The advantage of these updates is that collaborators can focus on a key part of computations without considering the other classes. For example, if one wants to improve the computational speed by nicely handling the response function, one should only create a new class of the data loader and override methods related to the response handling, like expected count calculation. Each base class has methods that must be overridden in subclasses. Using only these methods, ImageDeconvoltuion can perform image deconvolution without knowing the details of the data itself.

This update also includes the following changes:
- The image deconvolution can be performed using different datasets. For example, if the response function is changed for some reason, e.g., a part of the Ge detector becomes out of work during satellite operation, we need to use a different response function before/after such a change. Now, the code can handle multiple datasets and perform the image deconvoltion using them simultaneously.
- The optimization of normalizations of multiple background components. The background models should be set to the dataloader as a Python dictionary. The RL algorithm can optimize their normalization by referring to keys in the background dictionary, even using multiple datasets.

Screenshot 2024-06-11 at 16 46 19
Screenshot 2024-06-11 at 16 46 29
Screenshot 2024-06-11 at 16 47 01


hiyoneda:
I am going to update the notebooks related to the image deconvolution.

hiyoneda:
I've updated the notebooks. Also, in this PR, I replaced `print` with `logger.info` regarding the issue #48.

israelmcmc:
Thanks for working on this, @hiyoneda! I haven't gone through the code yet, just the description. Here are some initial thoughts and questions:
1. I like the idea of separating the different parts of the code. It will be very helpful both to optimize it and to add new features.
2. Instead of the exposure map, can we have the SpacecraftFile as an attribute of the base DataLoader class? It is more general, and the exposure map can be computed on the fly from there if needed. Some algorithms won't have exposure as a function of the model space --i.e. as a function of RA/Dec.
3. Note that in order for the codecov check to pass, you need to have at least 50% of the new code tested. It says "target 100%", but I currently set a 50% tolerance while we get used to it.

hiyoneda:
I attached figures to explain about the correspondence between the RL algorithm and the new functions.

Screenshot 2024-06-24 at 14 47 07
Screenshot 2024-06-24 at 14 47 16

israelmcmc:
Thanks for the figure, @hirokiyoneda. It's clearer now. On second thought, I agree it's fine to have the exposure_map in the base class. Most of the time we'll need it, and if for some reason you don't, we can replace it with a property that constructs and returns a vector on the fly.

hiyoneda:
I renamed `DataLoader` to `DataDeconvolutionAlgorithmInterface`. 
The subclass will be named `DataIF_AAA`. For instance, one corresponding to DC2 is named as `DataIF_COSI_DC2`.

hiyoneda:
I changed the class name to `ImageDeconvolutionDataInterfaceBase`. Sorry if it is confusing...

hiyoneda:
@israelmcmc @ckarwin 
I am preparing unit_test for these new classes. I wanted to use `data_pl.h5`/`bkg_pl.h5`, but their bins differ from the response `test_full_detector_response.h5`. For example, their energy axes are:
        event      : [ 100.  200.  500. 1000. 2000. 5000.] keV
        response : [ 150.  220.  325.  480.  520.  765. 1120. 1650. 2350. 3450. 5000.]

Can I replace `data_pl.h5`/`bkg_pl.h5` with new ones with the same axes as the response?

israelmcmc:
@hiyoneda That seems fine. As far as I can tell, none of the current tests or tutorials use that file.

hiyoneda:
The unit test has been added for the image deconvolution. It covers most of the codes (more than 90% for all).
@israelmcmc I think that now it is ready to be reviewed again. Can you check it?

```
cosipy/image_deconvolution/RichardsonLucy.py                              124      5    96%
cosipy/image_deconvolution/RichardsonLucySimple.py                         53      3    94%
cosipy/image_deconvolution/__init__.py                                     10      0   100%
cosipy/image_deconvolution/allskyimage.py                                  49      0   100%
cosipy/image_deconvolution/coordsys_conversion_matrix.py                   92      0   100%
cosipy/image_deconvolution/dataIF_COSI_DC2.py                             129     12    91%
cosipy/image_deconvolution/deconvolution_algorithm_base.py                 84      0   100%
cosipy/image_deconvolution/exposure_table.py                              161      4    98%
cosipy/image_deconvolution/image_deconvolution.py                          99      4    96%
cosipy/image_deconvolution/image_deconvolution_data_interface_base.py      40      0   100%
cosipy/image_deconvolution/model_base.py                                   21      0   100%
```

israelmcmc:
Thanks @hiyoneda. Something's up with codecov. It's not updating the coverage, I'll need to check it.

israelmcmc:
I'm closing and reopening this PR to kick codecov. There was a recent change in GitHub that broke it, but I think it's fixed now.

israelmcmc:
It took me a while, but I finally went through this code. Good job @hiyoneda! I think these changes will be very fruitful.

Here are my list of comments:
- The overall the structure is sensible. It will allow us to expand in multiple directions, but it's simple enough that I was able to create this [basic 1D toy test,](https://nbviewer.org/gist/israelmcmc/41fad3c5bef122fb3981c44fde28a867) something I'd been wanting to do for a while. It works! I didn't run the full tutorial, but I think this was a good check, along with your unit tests.
- I had to make some simple changes to make your code work without units. Please review them on my [PR to your PR]( https://github.com/hiyoneda/cosipy/pull/1)
- I checked that we can use this structure for the parallelization @avalluvan is doing on #189. @avalluvan: you basically need to put your code into the calc_expectation() and calc_T_product(). Either for one of the existing classes or a new ImageDeconvolutionDataInterfaceBase child class, I'm not sure what's easier. Also, I think it would be better to easier to disregard internally the background interface and make the background normalization part of the model, as described on slide 7 here [20240222-RLunbinned.pdf](https://github.com/user-attachments/files/16254208/20240222-RLunbinned.pdf). Both designs are possible with the current code though, whatever is easier.
- I also checked that this can work for the unbinned analysis (#222). For this, `ImageDeconvolutionDataInterfaceBase._event` needs to be set to 1 (a scalar is fine), `calc_expectation` and `_bkg_models` need to be calculated differentially (per phase space unit), and the likelihood need to be the one described on slide 6 here [20230817-UnbinnedCont-Israel.pdf](https://github.com/user-attachments/files/16254207/20230817-UnbinnedCont-Israel.pdf). The `_exposure_map` needs to take into account the full phase space, whether we had events or not, so it's good that it was decoupled from the response matrix.
- The documentation was not compiling. Partially because of an issue with the develop branch, fixed by 025c857d34a. After syncing with develop, it was still not compiling because the default of some functions were using astropy units. When we build the documentation, sphinx reads the definition of the function, but astropy is not yet installed at that step, so it failed to do any operations. On my [PR to your PR]( https://github.com/hiyoneda/cosipy/pull/1) I added some ways to address this, you can choose or try something else.
- On my [PR to your PR]( https://github.com/hiyoneda/cosipy/pull/1), I also added the images you posted here to the API documentation. I found them very useful and didn't want them to get lost. You can do a better job at it with the latex code, but you don't have to.
- I noticed that the tutorial is still mentioning "DataLoader", but that name has been deprecated.
- Small thing: I noticed a bunch of "this will be removed in the future" messages. It would be good to add a real [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) as opposed to using `logger.warn` or a comment in the code.

In summary, I'm pretty happy with this PR and I think it's almost ready to merge, with only a few small changes.

hiyoneda:
Thanks a lot for your reviews. Following them, I modified the codes as follows:

- Use DeprecationWarning instead of logger.warning for DataIF_COSI_DC2._modify_axes().
- Updated the notebooks for image deconvolution
  -  Replace "dataloader" with "data_interface"
  -  Use the data analysis in the galactic CDS as an example on the documentation webpage.

Also, I merged @israelmcmc's pull request to this branch. I hope that it is ready to merge now.

israelmcmc (approved):
All looks good to me, thanks @hiyoneda! Merging.

Issue #119, opened 2023-12-01
Need to optimize combine_unbinned_data method


ckarwin:
This method is using more memory than needed. Needs to be optimized.

ckarwin:
@krishnatejavedula I wanted to try to provide a bit of guidance on this issue. 

You'll see that the method reads in one dictionary at a time, and concatenates them to the final combined dictionary. In order to concatenate a given dictionary, it needs to be read into memory. However, it should be immediately removed from memory after being concatenated. I'm not sure that this was actually happening, b/c when we were testing it we saw that the RAM was multiplying with the number of combined files (beyond what we expect for just the new combined dictionary). So that's one thing to check. Although this may have already been resolved since opening the issue -- I'm not quite sure. 

At most, I think the RAM will need to hold the data for all combined dictionaries, plus the current dictionary that is being concatenated. One possible option to reduce this might be to try a memory map: https://numpy.org/doc/stable/reference/generated/numpy.memmap.html.

krishnatejavedula:
@ckarwin Thanks for the guidance. It does seem like the RAM usage is higher than expected and I have encountered this issue while working with Data IO during DC2. I’ll check on this and see if using NumPy's memmap can help reduce memory consumption.

I appreciate your input and will keep you updated.

israelmcmc:
See also #344

PR #190, opened 2024-06-14
Refactored combine_unbinned_data to use memory-mapped files

krishnatejavedula:
- Introduced temporary directory for memory-mapped files to handle large data efficiently.
- Used np.memmap for memory management to avoid excessive memory usage.
- Improved handling for different file types (HDF5 and FITS).
- Enhanced memory efficiency by expanding memory-mapped files dynamically.
- Added clear dereferencing of unused memory to aid garbage collection.

Cloese issue #119


cosipy/data_io/UnBinnedData.py
```diff
@@ -655,29 +656,61 @@ def combine_unbinned_data(self, input_files, output_name=None):
...
-            gc.collect()
 
-        # Write unbinned data to file (either fits or hdf5):
-        if output_name != None:
-            self.write_unbinned_output(output_name)
+        with tempfile.TemporaryDirectory() as tmpdir:
+            for each in input_files:
+                logger.info("adding %s..." % each)
+
+                # Read dict from hdf5 or fits:
```
ckarwin, line 664: Hi @krishnatejavedula, Can you please use `this_dict = self.get_dict(each)` here to replace lines 665 - 668. See my line 663 that you deleted.

krishnatejavedula: Hello @ckarwin, sorry about that. I have made the changes you've requested.

ckarwin:
Hi @krishnatejavedula I looked over your changes and ran some test in order to profile the memory usage. I tested combining the crab+crab, as well as the crab+total_background from DC2. The memory usage vs. time are shown below: 

memory_profile_1

memory_profile_2


These were obtained using memory_profiler, described [here](https://github.com/pythonprofilers/memory_profiler). As an additional sanity check, I also got the peak RAM usage directly from the linux command line (see method 4 [here](https://github.com/ckarwin/my_notebooks/blob/main/profile/mini_DC2_GRB/executable.txt)), which was consistent with the results shown here. 

So it looks like your changes do not really improve things very much. Do you want to check this on your machine to make sure I'm not missing something? One possible thing to try is as follows:
1. The purpose of the memory map is to utilize the disk memory instead of the RAM. But I don't think this is happening. Would it help to flush the memory maps, as described [here](https://numpy.org/doc/stable/reference/generated/numpy.memmap.html), and then manually delete them from memory?

If the improvement is in fact minimal, as it appears, I think it would be better to not implement these changes. Specifically, the use of memory maps is more complicated than the previous method, and this has a higher risk of potentially causing problems down the road.

krishnatejavedula:
Hi @israelmcmc, @ckarwin  I was working on this and tried to see if I could get better results. But I believe it’s better to close the PR for now and try it again later with a different approach.

israelmcmc:
Sounds good, thanks!

hiyoneda:
@krishnatejavedula @ckarwin, I was checking the memmap functionality in my environment. I did not use COSIpy; I purely used astropy.io.fits and monitored memory usage. 
I've noticed that it worked well when I loaded an unzipped fits file, but the memory usage exploded when using a **gzipped** fits file. So, perhaps, a zipped file may have to be unzipped internally, which causes huge memory usage or prevents the memmap from working correctly. Can you also check if it happens in your environment?

israelmcmc:
As per our discussion, I'm closing this issue. Feel free to keep using it to document attempts, or reopen it if something promising comes up.

PR #227, opened 2024-07-24
Fix docs. Add yayc to mocked imports

israelmcmc:
Broke here: 8b91ef04c766


israelmcmc:
I'm merging this directly since it's a small maintenance chance.

Issue #208, opened 2024-07-16
Routine to convert PA from one convention to another


israelmcmc:

israelmcmc:
Closed by #232

PR #232, opened 2024-08-12
Add polarization module with utility functions for polarization transformations

krishnatejavedula:
- Added polarization module
- Included util.py with functions for generating meshgrids, projections, and angle transformations
- Created __init__.py to expose key functions from the util module

Addresses issue #208


israelmcmc:
Thanks for working on this @krishnatejavedula! A few things:
1. I think that the source direction [x,y,z] and the ref_vector should be input by the user as SkyCoord object directly, as opposed to input individual coordinates and then transforming inside. Since a SkyCoord also contains the reference frame, you don't need to have a special function to transform between two conventions in SC coordinate to between internal and SC, just one single general function. It's also less error-prone and concise. 
2. I'm not sure that transform_pa_sc_to_celestial and transform_pa_celestial_to_sc are working correctly. To do this you need to transform the the reference vectors and source direction to the same frame, and then use transform_pa. The PA is not the same as RA. If you do 1) first, then you don't even need these functions though.
3. There should be a ref_vector for each orthographic convention. The stereographic convention doesn't need a ref_vector the way it's currently defined. It would be better to make a PolarizationConvention class, and then OrtographicPolarizationConvention and StereographicPolarizationConvention child classes that the user can create.
4. Once you are done, can you please add the corresponding unit tests? [Here's](https://github.com/cositools/cosipy/blob/develop/docs/dev/unit_tests.rst) how to do it. Or you can use the [tests for other modules](https://github.com/cositools/cosipy/tree/develop/tests) as examples.

eneights:
This looks really good @krishnatejavedula! Just a few comments:
1. I think it would be better for the functions to be able to input and output angles with arbitrary units, instead of only in radians. Maybe look into astropy's Angle object (https://docs.astropy.org/en/stable/api/astropy.coordinates.Angle.html)? 
2. As written, the orthographic projection equations are true only when the source direction vector is a unit vector. I think this should always be the case, but it might be better to have the equations be correct regardless of the length of the vector (divide the second term of the equations on lines 55-57 by the length of the source direction vector squared).
3. I'm not sure I'm understanding the project() function that just returns None (line 15 of util.py). Everything seems to run the same if I get rid of it; is there a reason it needs to be there?

krishnatejavedula:
Hi @eneights. Thanks for reviewing the code.

1. You are right. I will look into incorporating a more general system for inputs and outputs for the angles.
2. I see what you mean. I can add it to the code.
3. The project() function was just a template method that is supposed to be implemented in the rest of the subclasses. It can be removed entirely or left there as a place holder (possibly raising a NotImplementedError). It won't impact the functionality of the code either way.

eneights:
Thanks @krishnatejavedula for making those changes! 

Small correction: I think lines 55-57 need to be divided by norm_source**2.

Also, the polarization angle is typically defined between 0 & 180 degrees, so it would be preferable if pa_transformation() output an angle between 0 & pi instead of -pi & pi.

israelmcmc:
Hi @krishnatejavedula . Thanks again for working on this. I submitted a [pull request](https://github.com/krishnatejavedula/cosipy/pull/1) to your PR requesting what seems like a bunch of modifications, but most of this is code refactoring. The PA transformation you wrote is essentially what you wrote, just in different places. The main goal was to create a PolarizationAngle class that feels like SkyCoord's SkyCoord. It contains all the necessary inputs to perform the transformation: source direction and convention. In this case, PolarizationConvention is equivalent to astropy's BaseCoordinateFrame. I think this is less error-prone, as it is hard for the user to track all of these things by themselves.

Besides refactoring, there was one non-trivial change. I think there was an error in your calculation of px,py on the orthogonal convention. You had something like

```
px = ref_vector - (source direction . ref_vector) * source_direction
```

However, this is missing something since, by definition

```
ref_vector = (source direction . ref_vector) * source_direction + (source direction . px) * px
```

Your code was giving me something like this:

ref_plusZ_old

Some other changes:
- Your polatization/__init__.py was missing a `_` on both sides (it needs 2 underscores on both sides)
- Added the [documentation](https://github.com/israelmcmc/cosipy/blob/PR232_Israel/docs/api/polarization.rst) on conventions and conversion.
- Updated your unit tests accordingly.
- Users can now specify whether the convention is clockwise or counter-clockwise
- I added some special conventions that the user can call by name: IAU, and RelativeX/Y/Z. I believe this matches the IAU and MEGAlib's convention, but we should double-check this.
- The PA transformation accounts for different coordinate frames (e.g. galactic to SC), in addition to different conventions. This is achieved by transforming the polarization vector before moving to bare cartesian coordinates. 

@krishnatejavedula  and @eneights : Please take a look at let me know if everything still works the way it's supposed to, or if I broke something.

israelmcmc:
Thanks @krishnatejavedula. I’ll just wait for @zoglauer for confirm the RelativeX/Y/Z convention and then I’ll merge it

israelmcmc (approved):
I resolved some conflicts, and now it is ready to merge.

PR #233, opened 2024-08-12
Change print() to logging.info()

krishnatejavedula:
Replaced print statements with logging in various files to improve logging consistency, clarity, and debugging capabilities.
Updated logging statements in the following files:

cosipy/ts_map/fast_ts_fit.py
cosipy/ts_map/TSMap.py
cosipy/response/FullDetectorResponse.py
cosipy/spacecraftfile/SpacecraftFile.py

Addresses issue #48

### DIFF
--- a/cosipy/response/FullDetectorResponse.py
+++ b/cosipy/response/FullDetectorResponse.py
@@ -298,7 +298,7 @@ class FullDetectorResponse(HealpixBase):
 
             # If fast method fails, use long method, which should work in all cases.
             except:
-                logger.error("Initial attempt failed.")
+                logger.info("Initial attempt failed.")
                 logger.info("Using long method...")
                 nlines = sum(1 for _ in gzip.open(filename,"rt"))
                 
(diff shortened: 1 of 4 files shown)


cosipy/response/FullDetectorResponse.py
```diff
@@ -298,8 +298,8 @@ def _open_rsp(cls, filename, Spectrumfile=None,norm="Linear" ,single_pixel = Fal
 
             # If fast method fails, use long method, which should work in all cases.
             except:
-                print("Initial attempt failed.")
-                print("Using long method...")
+                logger.error("Initial attempt failed.")
```
israelmcmc, line 301: I think this should be info(). Despite the message, in this context, the algorithm is not really failing, it just couldn't do the fast method, which is fine. The problem with using error is that it will show up if the use decides to use the "error" logging level.

israelmcmc (changes requested):
Thanks @krishnatejavedula. I can merge it after we address the small change I suggested. I'll waive the patch coverage threshold requirement since this code is not adding new functionality.

israelmcmc:
Thanks @krishnatejavedula! I'm merging this

PR #234, opened 2024-08-21
Source injector for `cosipy`

Yong2Sheng:
This PR adds the source injector module for cosipy.

# Current functionality
1. Inject a point source module to expected counts and save it to a file for analysis. 
2. The point source model can be an analytic function defined from `astromodels` or a tabular function read from a spectrum file that meets the definition and format for MEGAlib simulations  
3. The tutorial notebook is at `./docs/tutorials/source_injector/Point_source_injector.ipynb`
4. The unit coverage is 100%.

# Something to improve in the future
1. The reading of the spectrum file should be flexible instead of limited to the MEGAlib format.
2. The input tabular function should support `energy` and `flux` that are astropy Quantity object (so that we can check the units for `energy` and `flux`).


israelmcmc:
I'm merging this in preparation for tomorrow. @ckierans run it with no problems.

PR #240, opened 2024-09-03
ASAD method for polarization

eneights:
This includes a class for polarization fitting using the azimuthal scattering angle distribution (ASAD) method, and an example notebook which fits the polarization of a GRB. It's currently able to handle a point source which is stationary with respect to the detector, and a stationary spacecraft orientation.


israelmcmc:
Thanks, @eneights! I haven't had time to review it yet, but I just want to add a comment to remind us to merge #232 first (a dependency to this PR)

israelmcmc (changes requested):
I haven't had time to review it yet, but I just want to add a comment to remind us to merge [#232](https://github.com/cositools/cosipy/pull/232) first (a dependency to this PR)

nmik:
I tested the jupyter-notebook and was able to run it through the end.

israelmcmc:
Closing and reopening this PR to update the list of files changed after merging #232

israelmcmc:
Hi @eneights. Thanks for working on this. I finally went through all of your code, sorry for taking so long. In summary, this is working fine, but I do want to request a couple of changes for FullDetectorResponse before merging. 
- The `polarization` flag passed by the user is not really needed, since we can get that information from the file itself
- You added large `if` blocks based on whether the response had polarization information or not, however, these are for the most part duplicated code. This adds to the same problem with the `sparse` flag, creating a big piece of code hard to maintain.

I created a[ PR to your PR](https://github.com/eneights/cosipy/pull/1) implementing the changes above. Please check that it still works as you intended, and if so, once merged with your branch we can merge this PR.

I do have other comments for `polarization_asad.py`. However, since these are self-contained and I want us to focus on the polarization analysis using 3ML, I don't think they need to be addressed before merging this PR. Here they are:

- About `calculate_uncertainties`. The ASAD method only really works in the Gaussian regime. I think this can be replaced by a simple sqrt(counts) (in-place when needed). Or, if you decide to keep it, it might be better to hide it as a local utility function (not in __initi__). Since it is not of general applicability, keeping it local saves you from having to maintain it.
- It would be better to store your ASADs as histpy Histogram objects, instead of the dicts. That way you have access to all the methods (e.g. `fill()`, binary operation `-`, `plot()`, etc) that should make your code easier to read and maintain. In addition, related to the point aboe, you can use histpy Histogram the bin_error to get the uncertainties.
- You are computing the error for the ASAD prediction that come from the response the same way as you do for data (sqrt(counts)). This is not really correct. The errors for the response come from the triggered counts in simulations, not from the predicted counts after convolutions. We don't currently keep track of these, but since we use many many simulated events, these should be way smaller than what you are estimating. It might be better to just exclude these errors from the error propagation and just keep the errors for the data.
- I would prefer if `calculate_azimuthal_scattering_angle` becomes part of the PolarizationAngle class since this is of general use, not just for the ASAD method.
- I know this was just a proof of principle, but many parts of your code can be optimized a lot. It would be good to run a profiler. Some of the things I noticed 1) sometimes you use `project` just to get the number of bins 2) use `for` loops for operations that can be parallelized 3) Repeat operation within a loop that could be cached
- The fit to the `constant()` function can be replace by just `np.mean()`, doesn't it?
- If not intuitive when function that are meant to perform a computation also generate plots by default.
- We should be using `logging` instead of print statements.

We could open an issue with these for future reference if you don't plan to continue working on the ASAD method for now.

israelmcmc:
I think the tests failing are related to the new ori format after this PR was created:
https://github.com/cositools/cosipy/issues/263

eneights:
Thanks @israelmcmc! Your changes to FullDetectorResponse look good to me. And I agree with all of your feedback; I'll make those changes when I start working on this. I opened issue #275 to keep track of these.

israelmcmc:
@eneights Thanks for merging my suggestion. It seems the tests are no longer failing due to the ori file, but now they are failing on a test related to this PR:
```
FAILED tests/polarization/test_polarization_asad.py::test_convolve_spectrum - assert 6 == 7
 +  where 6 = <histpy.histogram.Histogram object at 0x7efc189660e0>.ndim
 +    where <histpy.histogram.Histogram object at 0x7efc189660e0> = <cosipy.polarization.polarization_asad.PolarizationASAD object at 0x7efc18eb65f0>._expectation
FAILED tests/polarization/test_polarization_asad.py::test_polarization_fit - astropy.units.core.UnitConversionError: Can only apply 'less_equal' function to dimensionless quantities when other argument is not a quantity (unless the latter is all zero/infinity/nan).
```

Can you take a look, please?

israelmcmc:
Thanks for fixing the test @eneights and @hirokiyoneda. I'll merge this. @hirokiyoneda I know you saw weird units with an older version, but Eliza did not. If you still see that issue after merging this please let us know and we’ll revisit it.

Issue #194, opened 2024-06-28
Revisit units of extended source fit tutorial


israelmcmc:
https://github.com/cositools/cosipy/blob/develop/docs/tutorials/spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb

The tutorial says that the units of the K normalization are differential:
[``` K          = [0.00046, 0.0011, 0.0027, 4.5e-3]/u.cm/u.cm/u.s/u.keV```](https://github.com/cositools/cosipy/blob/65706f739ea90e0d61103599908401e8d8f923d7/docs/tutorials/spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb#L2611)

However, the code normalizes the spectrum based on the integral flux (assuming a 1 keV bin size)
https://github.com/cositools/cosipy/blob/65706f739ea90e0d61103599908401e8d8f923d7/cosipy/threeml/custom_functions.py#L177

In addition, on #191 we're changing the definition of K to be unitless. For now, we changed the file in order to get the same numerical output, but we still need to fix this.

israelmcmc:
Closed by #241

PR #241, opened 2024-09-08
Revised units for parameter K (ph/cm2/s) in extended source fit tutorial to match the units in the SpecFromDat file

krishnatejavedula:
Addresses issue #194


israelmcmc:
Hi @krishnatejavedula, thanks for working on this. First, I verified that you are changing only this line
```
-    "K          = [0.00046, 0.0011, 0.0027, 4.5e-3]/u.cm/u.cm/u.s/u.keV\n",
+    "K          = [0.00046, 0.0011, 0.0027, 4.5e-3]/u.cm/u.cm/u.s\n",
```
the rest are just images in the notebook that are actually the same, as expected. 

I think there's something still off. Looking at [SpecFromDat](https://github.com/cositools/cosipy/blob/c87577d35a213f7a364e6cd2d428f7c46668ad09/cosipy/threeml/custom_functions.py#L99) the `K` should indeed be integral flux --i.e. 1/cm2/s, since we perform this normalization operation:
```python
dataFlux = dataFlux  / np.sum(dataFlux * ewidths)
```
However, the `evaluate()` should still have units of flux density (per unit of energy) but this is not what happens. For example:
```python
from cosipy.threeml.custom_functions import  SpecFromDat
import astropy.units as u

spec = SpecFromDat(dat="</path/to/cosipy>/docs/tutorials/spectral_fits/extended_source_fit/OPsSpectrum.dat")

K = 1/u.cm/u.cm/u.s

spec.evaluate([1]*u.MeV, K)
```
returns `[7.35916986e-06] 1 / (cm2 s)`. 

The issue is that when you divide by `ewidths` the units should also be included. As per the `.dat` documentation, `ewidths` has units of keV, the missing unit. In summary, I agree with your change in the notebook, but this should be accompanied by the appropriate change in `SpecFromDat`. Overall, although the result will be identical as it was before, the units and meaning of each variable will make sense.

I'm actually not sure how this will affect other code using `SpecFromData`. I think it's mostly the SourceInjector, so you should check. I tried to check the [SourceInjector tutorial](https://github.com/cositools/cosipy/blob/develop/docs/tutorials/source_injector/Point_source_injector.ipynb) but it doesn't have units!:
```
spectrum = SpecFromDat(K=1/18, dat="crab_spec.dat")
```
I think that should be revisited as well.

israelmcmc:
Closing and reopening to re-run the tests.

israelmcmc:
@krishnatejavedula I'm getting this error 
```
UnitConversionError: '1 / (s cm2)' (particle flux) and '1 / (keV s cm2)' are not convertible
```
when executing the extended source notebook, here:
```
Cell In[34], line 6
      4 c = SkyCoord(l=0*u.deg, b=0*u.deg, frame='galactic')
      5 c_icrs = c.transform_to('icrs')
----> 6 ModelCentralPoint = PointSource('centralPoint', ra = c_icrs.ra.deg, dec = c_icrs.dec.deg, spectral_shape=SpecCentralPoint)
      7 ModelNarrowBulge = ExtendedSource('narrowBulge',spectral_shape=SpecNarrowBulge,spatial_shape=MapNarrowBulge)
      8 ModelBroadBulge = ExtendedSource('broadBulge',spectral_shape=SpecBroadBulge,spatial_shape=MapBroadBulge)
```

Can you please try to run it again after syncing with develop? Maybe it's something that changed since you opened this PR (sorry for just getting to it!)

krishnatejavedula:
Hi @israelmcmc, it's okay. I am checking it now.

krishnatejavedula:
Hello @israelmcmc.

 I have reverted the changes I made in this PR and updated the docstring for `SpecFromDat` so that the units of `K` match those used in the notebooks. 

We may need to revisit the units of `K` in the future, but for now, this resolves the issues we were encountering.

Yong2Sheng:
This PR conflicts with the merged [PR399](https://github.com/cositools/cosipy/pull/399), which removes the `log10` transformation on parameter **K** and updates its bounds to a linear range (0.0–1e6) and delta to 1. These updates are not present in this PR, so the two PRs currently conflict.

Since PR399 is newer, I resolve the conflict by adopting the PR399 changes for consistency:
- remove the `transformation` as done in PR399  
- set `min = 0`, `max = 1e6`, and `delta = 1` following PR399  
- keep the units `ph/cm²/s/keV` as defined in this PR

Issue #226, opened 2024-07-22
Unit tests


israelmcmc:
Umbrella issue to keep track of the unit test coverage progress. While the eventual goal is to have 100% coverage, for now I'll check each module when it reaches 80%. 

- [ ] Imaging. 97% coverage (When #188 is merged) @hirokiyoneda.
- [ ] Threeml. 28% coverage. @eneights 
- [ ] Response. 44% coverage. @israelmcmc (others can help...)
- [ ] SC file. 13% coverage. @saurabhmittal23 @Yong2Sheng 
- [ ] TS map. 21% coverage. @Yong2Sheng 
- [ ] Data IO. 84% coverage. @ckarwin 

See details here: https://app.codecov.io/gh/cositools/cosipy/tree/develop/cosipy

israelmcmc:
Updated progress after PR #231

augustus-thomas:
Hello @israelmcmc! I would like to write some unit tests to help this project get better coverage.
Are there any tests that I can be assigned? I see that there are a few PR's open that are untested.

israelmcmc:
Thanks @augustus-thomas! How about you add some tests for the response module to increase its coverage?

The tests are here:
https://github.com/cositools/cosipy/tree/develop/tests/response

You can see the lines that are not covered here:
https://app.codecov.io/github/cositools/cosipy/tree/develop/cosipy%2Fresponse

Let me know if you need more info to get you started :)

augustus-thomas:
Hi @israelmcmc thank you so much. I am very excited to finally start helping out!

It looks like a large number of the uncovered lines in `FullDetectorResponse.py` are due to the untested method `_open_rsp()` This requires an `rsp.gz` file in the test data folder, which we do not currently have. Do you know of a good endpoint I can fetch with `fetch_wasabi`? I am unfamiliar with generating or fetching a file like that, and we would need to push it to the  `test_data` directory to get more coverage!

Also, is there a way to browse the cosipy data that can be fetched with wasabi?

israelmcmc:
@augustus-thomas You can use the following command to browse wasabi (with your cosipy environment loaded):
```
[...]

Issue #82, opened 2023-10-02
Generalize getting the units from a 3ML/astromodels spectrum 


israelmcmc:
Right now we need a switch that looks for the specific spectrum type. There should be a better way that works with all functions, including new ones in the future.

@lommler improved on the previous version, see #28

However, as it turned out during PR #78, other functions still need customized if statements 
https://github.com/cositools/cosipy/pull/78#discussion_r1337786622_

eneights:
This should now be fixed in PR [#78](https://github.com/cositools/cosipy/pull/78)

israelmcmc:
Cool! Looks much better! Thanks

israelmcmc:
It seems that there are still some function in astromodels that do not have an explicit normalization parameter, so the current solution, while it's progress, doesn't work on all cases. I'll leave this issue open in the meanwhile.

eneights:
I updated this in PR [#78](https://github.com/cositools/cosipy/pull/78), but it still doesn't work for all cases, since I wasn't sure about the units for a few of the models.

israelmcmc:
It seems that the solution to this issue is two-fold:
1. Fix how we initialize and set parameters in cosipy. Currently we always do something like:
```
from astromodels import Powerlaw
from astropy import units as u

fun = Powerlaw()

K = 1/u.MeV
fun.K.value = K.value
fun.K.unit = K.unit
```

We should instead use a call to `set_units` call. e.g.:
```
from astromodels import Powerlaw
from astropy import units as u

fun = Powerlaw()

fun.set_units(u.keV, 1/u.keV)

fun.K = 1/u.MeV
```

`set_units` is part of the definition of [Function1D](https://github.com/threeML/astromodels/blob/3515d4ca48dbd3f5fe2e4aa2c1fa5d9988da3e39/astromodels/functions/function.py#L1182), so all functions know how to handle their normalization parameter such that the output units match. It also makes setting the parameter values easier, since now they handle the units automatically.

2. Once we follow this procedure, we can call `y_unit`. In particular, [here](https://github.com/cositools/cosipy/blob/e3d4d834dfc324d147154b97c98fa2f0d1fae302/cosipy/response/functions.py#L89) we can replace `get_spectrum_unit` (the offending function).

This was suggested by @PreisTo in https://github.com/threeML/astromodels/issues/280

PR #243, opened 2024-09-19
Unit Tests for PointSourceResponse

augustus-thomas:
Partially addresses response testing for Issue #226 

- `PointSourceReponse.py` now has its own associated test file called `test_point_source_response.py`.
- Many spectra model types are tested in the `get_expectation` method call, which should ensure full coverage.
- The point source response used is generated from a call to `FullDetectorResponse.get_point_source_response()`
- The `FullDetectorResponse` used was generated from the `test_full_detector_response.h5`

Let me know if this PR is helpful! I am very excited to contribute something to this team :)


israelmcmc:
Thanks @augustus-thomas! This looks good to me. One suggestion: currently you are basically testing whether various methods run without crashing, which is good. In addition, it would be better to also test that the output is what's expected. For example, you can add a check (an assert) that the total expectation equals (or approximately equals) what we're currently getting, so that if a future version of the code breaks this it won't go unnoticed. 

Please let me know if you have time and are willing to do this. I understand if you don't, in which case I would merge this anyway since it's nonetheless progress!

augustus-thomas:
Hey! Thanks for the feedback here @israelmcmc. I would like to insure that the expectations are correct, and I do have the time to issue a new commit for that.

I will follow the example in `test_full_detector_response` for the assert statements. I have been testing the values of the `np.ndarray` given when invoking `expectation.contents.todense()`. It turns out that these are filled with zeros! **I think the spectra need to be initialized in some non-trivial way, but I am not sure yet.**

israelmcmc:
Sound good, thanks @augustus-thomas . You can see an example of how to initialize a model and get the expectation here: https://cositools.github.io/cosipy/tutorials/response/DetectorResponse.html

augustus-thomas:
This is now updated based on our conversations. The forced push was because I included some other commits on accident

israelmcmc:
I'll close and reopen this PR to kick the unit tests, which seem stuck.

israelmcmc:
@augustus-thomas Thanks for the changes. It seems something is off with the unit test 🤔 . Please take a look at the "checks" tab. I tried to reproduce this in my system, and they also fail at that point, but due to a different reason: 
```
   def test_get_expectation():
        # supported spectral functions
        ## see astromodels.functions.function.Function1D.[function_name]()
        ## 3 sig-figs
>       assert 631 < np.sum(psr.get_expectation(Constant(k=1e-1)).contents) < 632
E       AssertionError: assert 219126308568.4652 < 632

```
🤔 

Have you seen this?

israelmcmc:
Closing and reopening to re-run the tests.

augustus-thomas:
I have managed to make some progress here. My issue was that the spectrum model parameters were dimensionless, so the expectation counts had dimensions `cm2 s keV` after convolving with the PSR. The expectations should have been dimensionless. I fixed that by normalizing the spectrum models and giving their fit parameters units `1/(u.keV * u.cm**2 * u.s)`

Normalization was straightforward for the basic models (Constant, Line, Cubic, Quartic, Step) but is less obvious for the other ones. For example, how to change the units of the Photometric absorption model?

israelmcmc:
Thanks @augustus-thomas. I made a commit to your branch to resolve a small merging conflict. That will allow the tests to run, let's see how that goes.

Ah yes, the normalization unit for the different spectra are sometimes not obvious. That's been a long issue that prevents us from making some stuff automatic. I think you are referring to #82

augustus-thomas:
1. It seems the tests are passing on my machine and failing on the CI/CD pipeline, which I need to address. Investigating further I am finding that the expectations generated by `main` and `develop` branches are significantly different, even with the same code and the same models used. The exposure map timeframe is from 1999 to 2010. The difference might be caused by a change in the full detector response file. It looks like the `test_full_detector_response.h5` file was changed in `develop` in the May optimization changes from @jdbuhler . We might want to validate the correctness of the new response. For example, I am getting very negative expected counts, which are non-physical. I don't see any other issues relating to unexpected expectations with the new response though. 
2. I had initially planned to test every model so as to test all of the `else-if` branches for each model type in `functions.py`After checking the test coverage on my branch, it looks like most of these models are covered! I am going to focus on getting `PointSourceResponse.py` up to `100%` coverage

augustus-thomas:
Some notes: 
To get to full coverage I will need a PSR or FullDetectorResponse .h5 file with a polarization axis. @israelmcmc do you know where I might find one?

augustus-thomas:
I am working on testing the `get_integrated_spectrum` with a GenericFunction model. I want to cover the `scipy.integrate` case. However, it looks like we are setting the `spectrum_unit` to `spectrum.K.unit` instead of `spectrum.k.unit` in `functions.py` I will change that typo in this PR.

augustus-thomas:
We cannot cover the `PhAbs()` spectrum case of `get_integrated_spectral_model` from `get_expectation` in the PSR class. This is because the units of the flux are set to `spectrum_unit` or `u.dimensionless` and so the expectation is not normalized and `psr.get_expectation` throws an error. To properly test that case I would suggest we write other `allskyimage` tests

augustus-thomas:
Short of a `test_response` with a `'Pol'` axis, this is as far as I can get!

augustus-thomas:
It looks like the "non-physical" expectation behavior in `get_expectation` is similar to that mentioned in Issue #363. It seems to result from the change of scale to `log10` in the new response version 2, rather than a scale problem on the histpy side. Since expectation is used for maximum log-likelihood algorithm, as @israelmcmc point out it is okay for it to be negative.

What is the best way for the user specify that they would like physical expectations? Should we add this to the documentation?

israelmcmc:
Hi @augustus-thomas. Sorry I didn't reply earlier. Thanks for working on this!  I see that the tests are working and the coverage is great. 

To your questions:
- You can use this response file to test polarization. I think it's also used in some other unit tests. https://github.com/cositools/cosipy/blob/develop/cosipy/test_data/test_polarization_response.h5
- I want to better understand the problem of negative values. On the one hand, as long as the flux normalization is positive, the PSR should never return a negative number of expected counts. On the other, it's OK for the _fitted_ value of the normalization to be negative, since the background will fluctuate around the mean --that's what I meant to say in #363. Can you clarify which one of these cases you are seeing? Or maybe neither and it's something else?

augustus-thomas:
>Sorry I didn't reply earlier. 

No worries at all. I see there are interfaces to refactor and other important open PRs. I suspect COSI is very busy leading up to the COSIfest.

>You can use this response file to test polarization. I think it's also used in some other unit tests. https://github.com/cositools/cosipy/blob/develop/cosipy/test_data/test_polarization_response.h5

Thanks so much for pointing this out! I had not seen that file :)) I committed the tests for the polarization response and we are at `100%` on PointSourceResponse now.

>I want to better understand the problem of negative values. On the one hand, as long as the flux normalization is positive, the PSR should never return a negative number of expected counts. On the other, it's OK for the _fitted_ value of the normalization to be negative, since the background will fluctuate around the mean --that's what I meant to say in [Unexpected behavior from custom function SpecFromDat #363](https://github.com/cositools/cosipy/issues/363). Can you clarify which one of these cases you are seeing? Or maybe neither and it's something else?

I will work more on plotting and verifying this. For now an example. In our expectation from the Quadratic spectrum hypothesis, expectation values corresponding to the edges `2511.89 keV`, `3981.07 keV`, and `6309.57 keV` `Em` bins are highly negative (`~-1e+12`). That is over all `Phi` and over all `PsiChi.` All expectation values in those three different 2D slices of the expectation are negative. My thinking was that perhaps they should be interpreted as being stored/encoded in `base10`? Perhaps the stored expectation values are `log10` of `n` counts. For example, an expectation value of `v = -1e+12` in a given bin means that `n = 10^(v) = 10^(-1*10^12)` for that bin? I don't know.

It is quite tedious to visualize 3D arrays for verification. Maybe voxel plotting with plotly? I hope to plot the meshes and also check negative values for the other hypotheses in the coming week. In the meantime here is a Python code in case you want to check it out.

```python3
import numpy as np
from histpy import Histogram, Axes, Axis
from cosipy import test_data
from cosipy.response import PointSourceResponse, FullDetectorResponse
from mhealpy import HealpixBase, HealpixMap
from scoords import SpacecraftFrame
from threeML import Quartic
import astropy.units as u
response_path = test_data.path/"test_full_detector_response.h5"
with FullDetectorResponse.open(response_path) as response:
    exposure_map = HealpixMap(base=response,
                                    unit=u.s,
                                    coordsys=SpacecraftFrame())

    ti = Time('1999-01-01T00:00:00.123456789')
    tf = Time('2010-01-01T00:00:00')
    dt = (tf-ti).to(u.s)

    exposure_map[:4] = dt/4
    psr = response.get_point_source_response(exposure_map = exposure_map)
norm = 1 / (u.keV * u.cm**2 * u.s)
quad = Quadratic(a=1e-1, b=-4e-5, c=1e-9)
quad.a.unit, quad.b.unit, quad.c.unit = norm, norm, norm
exp = psr.get_expectation(quad)
assert np.isclose(np.sum(exp.contents), -1.53393018e+12, rtol=1e-8) ##!!
arr = np.array(exp.contents)
for em_idx in range(0, arr.shape[0]):
    if np.all(arr[em_idx, :, :] < 0):
        print("Suspicious Energy Edge:" exp.axes[0].edges[em_idx] )
```
If you do end up running that let me know if it throws errors and I can troubleshoot it for you.

augustus-thomas:
In the above code `response` is a `cosipy.FullDetectorResponse` object, which inherits from `HealpixBase` class. `HealPixBase` allows for each `axis` in `HealPixBase.axes` to have an `axis_scale`. The `axis_scale`of `Ei` and `Em` is indeed `log` in the new response. According to my knowledge, `axis_scale` controls the embedding of values in the `HealpixBase` object. 

Our issue is somewhere in either `FullDetectorResponse.get_point_source_response()` or `PointSourceResponse.get_expectation()`.

### FullDetectorResponse.get_point_source_response()
https://github.com/cositools/cosipy/blob/e9dca56e9b808a1e42cd9f21da36d4d826bb0d72/cosipy/response/FullDetectorResponse.py#L449-L452
I don't know the requisite information to check if this method is correct when it is given a log scale in the energy axes of the FullDectectorResponse. However, I can see that `axis_scale` is not passed to the `PointSourceResponse(Histogram)` constructor here (there may be other places it is constructed in the control flow). If it is possible to initialize a Histogram with a given axis scale we could copy the scale from the full detector response. I am not sure if that would change the behavior of the Histogram though.

### PointSourceResponse.get_expectation()
https://github.com/cositools/cosipy/blob/e9dca56e9b808a1e42cd9f21da36d4d826bb0d72/cosipy/response/PointSourceResponse.py#L97-L101
Even if we copied the scale it is unclear if the result of this function would change. **Is the expectation the dot product of the flux and the response if both are embedded in `log10` in the contents array?**

The easiest short term solution is:
1. Change PointSourceResponse() construction to pass the `axis_scaling` from the `FullDetectorResponse` (and elsewhere?)
2. Enforce scaling checks and errors in PointSourceResponse.get_expectation() (and elsewhere?)
3. Write a manual convolution kernel for `log10` scaling so the expectation is correct. We could put that in `functions.py` because it will likely be useful elsewhere.

I am unsure what good long-term changes might include.

israelmcmc:
@augustus-thomas For this specific case of the `Quadratic` function, I think the negative expected counts are fine. Since you `b`, the spectrum (flux vs energy) will be negative --i.e. unphysical-- at some point. I haven't plotted it, but I suppose it's going negative within the COSI energy range, and therefore a negative number of expected counts is, well, expected. Please try changing the `b` values to something positive and let us know if that fixed it.

Btw, the `log` in Axis only determines the interpolation, in particular which value is considered the center of the bin (linear vs geometric center)

augustus-thomas:
Thank you for patiently explaining spectra and reading this long PR. It makes sense now why a negative "unphysical" spectrum would yield negative counts. Indeed **after changing the quadratic to a positive value function** we get positive value counts. I don't think it is required to plot it anymore. (for reference I changed it to a=1e-1, b=4e-5, c=1e-9).

I also changed all of the tests here to use positive value coefficients so that the spectra are physical and correct.

Also thank you for explaining the scaling is for interpolation rather than embedding.

israelmcmc:
No problem! Thank you for working on this.

It seems that all issues have been resolved and the tests look reasonable to me. I'm merging!

Issue #209, opened 2024-07-16
Polarization fit with Stokes parameters


israelmcmc:

israelmcmc:
Will be closed by #246.

israelmcmc:
Closed by #246

PR #246, opened 2024-10-02
Stokes parameters polarization measurement

nmik:
This script is the starting point to have a Stokes parameters formalism to compute the polarization from a source (with known location in the sky).

### DIFF
--- a/tests/polarization/test_polarization_stokes.py
+++ b/tests/polarization/test_polarization_stokes.py
@@ -12,7 +12,7 @@ from cosipy import test_data
 analysis = UnBinnedData(test_data.path / 'polarization_data.yaml')
 data = analysis.get_dict_from_hdf5(test_data.path / 'polarization_data.hdf5')
 response_path = test_data.path / 'test_polarization_response.h5'
-sc_orientation = SpacecraftFile.parse_from_file(test_data.path / 'polarization_ori.ori')
+sc_orientation = SpacecraftFile.open(test_data.path / 'polarization_ori.fits')
 attitude = sc_orientation.get_attitude()[0]
 
 a = 10. * u.keV


israelmcmc:
Closes #209

israelmcmc:
@eneights I think you were looking at this PR, is that correct?

eneights:
@israelmcmc Yes, I started looking at it and have needed to focus on other things, but it's on my to do list. Sorry for the delay!

israelmcmc:
Thanks for confirming @eneights. No worries, I just wanted to make sure the assignee and status label were correct.

eneights:
This looks good! I made some minor changes, which I'm creating a pull request for into Michela's branch once I figure out github. @nmik once that PR is merged into your branch, this PR will automatically update with those changes

For reference, this is how the results for the ASAD and Stokes methods compare:
Stokes: 81% +/- 5% at 87 +/- 2 deg with MDP = 16%
ASAD: 95% +/- 7% at 88 +/- 2 deg with MDP = 16%
Simulated: 80% at 90 deg

eneights:
@nmik I'm not sure what I did but I guess I pushed the changes into your branch. Can you take a look and make sure you're okay with them?

eneights:
The unit test was using an outdated response file. I switched it to the current polarization response in `test_data`, but the unit test is still failing

eneights:
Hi @nmik, can you please add a way to change the ASAD bin sizes/edges? This seems to currently always default to 20 bins, but should be adjustable. The unit test is still throwing an error because there are NaNs in the ASAD, which is due to there being too many bins for the coarse unit test response. Thank you!

nmik:
Hi, working on it right now. should be ready to pull in a few.
Here is the fix (it was quite easy to implement).

## Changes
I now use this definition in the __init__ method of class PolarizationStokes:
`        self._nbins = self._response.axes['Pol'].nbins`
`        print('Number of azimuthal angle bins used:', self._nbins)`

This variable self._nbins is now passed to 

- `create_polarized_asads()`,  and `create_unpolarized_asad()` in the method 'calculate_average_mu100' (which is only used to compute the modulation factor) and `simulate_unpolarized_stokes` (which is not used at the moment).

## Considerations
Final PD and PA predicted seems to be ok :

     PD: 84.51 +/- 5.47 %
     PA: 87.12 +/- 1.88 deg

The ASAD binning in not used to predict the Stokes-I, Stokes-U and Stokes-Q, so it doesn't change the measured values, but it does play a role in computing the modulation factor, which is used to compute the intrinsic polarization and the MDP.

## Diagnostic Plots
Some diagnostic plots are attached: everything looks good. 

image

image
Modulation PA bin 9
Modulation PA bin 10

Modulation PA bin 11

Modulation PA bin 0
Modulation PA bin 1
Modulation PA bin 2
Modulation PA bin 3
Modulation PA bin 4
Modulation PA bin 5
Modulation PA bin 6
Modulation PA bin 7
Modulation PA bin 8

-

eneights:
Thanks for making this change @nmik!

I still don't see a way for the user to specify the number of bins or bin edges, which is necessary to get the unit test to work with that extremely coarse response. There are currently empty azimuthal scattering angle bins, and it's necessary to use fewer bins in the ASAD than in the response. For example, I need to use 10 bins in my ASAD method unit test, while there are 12 in the unit test response. 

A plot is still being produced automatically in the unit test, with no option to disable it. I think it comes from line 458 in polarization_stokes.py. Could you please add a show_plots flag in __init__() for the user to choose whether to show the plots that are generated? Maybe this can also apply for line 483.

israelmcmc:
Nice to see this merged! Thank you @nmik and @eneights !

PR #252, opened 2024-10-20
Line background estimation

hiyoneda:
We implemented the first version of the line background estimation class. 
The basic idea is as follows:
- Extracting the event distribution in the Compton data space from adjacent energy bins of the energy of interest.
- Making a binned histogram using the extracted events.
- Estimate the total number of expected background counts in the line energy bin by fitting the adjacent energy bin data, and modify the normalization of the binned histogram accordingly.

We also added an example notebook, on which we generated a background model for Al-26 line. 

This class is very preliminary, and there are several areas for improvement. Future ideas include:
- We may add more options in the minuit fitting, e.g., limiting the parameter region, fixing some parameters.
- We may apply smoothing to the background histogram, which may help to mitigate Poisson fluctuation in the model.


ckarwin:
@hiyoneda I plan to check this first thing tomorrow. Do you plan to add unit tests?

ckarwin (approved):
Runs fine for me, and additions all look good.

PR #256, opened 2024-10-24
Add spectral fit unit test

eneights:
Add unit test for spectral fitting


ckarwin (approved):
Thanks for adding these tests, @eneights. Everything seems to be running fine.

PR #257, opened 2024-10-24
Add SpectrumFileProcessor class for automated spectral data conversion and formatting

krishnatejavedula:
- Implemented SpectrumFileProcessor class to handle energy and flux data conversion for spectrum files not in the MEGAlib spectrum file format.
- Added energy conversion feature: converts frequency (Hz) to energy (keV) and flux from ergs/cm²/sec to ph/cm²/sec/keV using astropy units.
- Implemented data filtering to restrict energy range to 100 keV - 10,000 keV for accurate source injector functionality.
- Reformatted output to ensure compatibility with SpecFromDat class and consistency across cosipy and DCs spectral files.


hiyoneda:
@krishnatejavedula Thanks for this work. Looks good. I have a quick comment. `threeml/custom_function.py`, to which you added this new class, is used to implement new spectral and spatial functions for astromodels in threeml fitting. I think that your new class can be used more generally. Can you think of a better directory for this class, maybe dataIO?

israelmcmc:
Thanks @krishnatejavedula. Any updates to this?

- I think what @hiyoneda suggested makes sense. Since there don't seem to be any interactions of SpectrumFileProcessor with 3ML, you can move this class to its own file inside the DataIO module.
- Besides that, the other thing to resolve is to add unit tests to cover this new class.

Yong2Sheng (approved):
I tested the module and it works well.

israelmcmc:
@krishnatejavedula What's the status of this PR? are the unit tests the only thing missing?

krishnatejavedula:
@israelmcmc It believe the code needs some refactoring and must be checked to see if it conflicts with any new changes to the source injector.

israelmcmc:
Got it, thanks for the update @krishnatejavedula!

PR #258, opened 2024-10-24
PSR with Earth occultation

ckarwin:
I modified the scatt map to include Earth occultation. The ori file now needs to include columns for altitude and Earth zenith (i.e. spacecraft location). Moreover, the scatt map should only be calculated for a single source position, and likewise for the point source response. Essentially, when filling the scatt map histogram, the code sets the weights to zero for pointings where the source is blocked by Earth. 

I added a simple example notebook in the response tutorial directory where I tested the changes. Everything seems to be working. As a simple test, I placed a point source at the Galactic center, and made an ori file where the instrument points to the Galactic center for 30 seconds. For the first 15 seconds the instrument is in front of Earth, and for the second 15.1 seconds it's behind earth. Summing the scatt map gives a total exposure time of 15.1 seconds, as expected.


ckarwin:
Ok, I fixed some bugs, and all checks have passed for unit tests.

GallegoSav (approved):

Issue #34, opened 2023-05-05
need to export the python path of cosipy in your bashrc file


GallegoSav:
If you install only cosipy from scratch (and not cositool) you need to add this line on your `bashrc` file : 
`export PYTHONPATH=$PYTHONPATH:/your/path/tocosipy/folder`

Otherwise you will get errors when trying to import cosipy functions

israelmcmc:
I think it is working in DC2. @GallegoSav if it doesn't work for you, please feel free to reopen this.

PR #259, opened 2024-10-29
MAP RichardsonLucy

hiyoneda:
I added a new image deconvolution algorithm, MAP Richardson-Lucy. It is based on Maximum A Posteriori (MAP) Estimation, and the RL algorithm is modified to include prior distributions flexibly.


israelmcmc:
@hirokiyoneda I have some cycles now to review this. Can I just ask you to resolve the conflict first, please? I think it's mostly the change in the tutorial names from #309. I also touched the tutorials in #334, but I think only the cells that used fet_wasabi_file.

hiyoneda:
@israelmcmc Yes, I merged the current develop branch and modified the 511 keV notebooks for now.

israelmcmc:
Thanks @hiyoneda! This looks good to me!

For reference

- I tested the 511keV-ScAtt-ImageDeconvolution.ipynb as is. It works and produces the same results as before.
- I modified `algorithm: "MAP_RL"` in the parfile. This is the difference of the results with the current output:

Screenshot 2025-04-11 at 9 24 37 AM
Screenshot 2025-04-11 at 9 24 45 AM

Which I think is expected since it's using the default values, which I think mean no priors, which recovers the behavior of the regular RL.

- I then modified added this to the parfile, which are the example values in the docstring of the `MAP_RichardsonLucy` class.
```
    prior:
      TSV:
        coefficient: 1.e+6
      gamma:
        model:
          theta:
            value: .inf
            unit: "cm-2 s-1 sr-1"
          k:
            value: 0.9
        background:
          theta:
            value: .inf
          k:
            value: 1.0
```

Screenshot 2025-04-11 at 9 44 50 AM
Screenshot 2025-04-11 at 9 45 15 AM

I'm not sure exactly how to interpret them, but I think it shows that in general, the code works. At least it's not exactly the same result, and it's still reasonable.

Everything looks good to me. I'm merging! I made a trivial change to your notebook so it can run automatically, so I'm merging #34 right after merging this.

For the future, I think it would be nice if RichardLucySimple were kept, well, simple 😬. That is, with the minimal set of parameters to make it work. You could add an intermediate class e.g. RLAugmented, deriving from RLSimple and adding these extra parameters, and RL and MAP_RL can derive from RLAugmented. Maybe other naming conventions would be better, but you get the point. This doesn't prevent this from being merged, just something for later, as RichardLucySimple can be the landing point for new students.

Btw, I see that you really despised "loglikehood" without the dash 😆

PR #276, opened 2025-01-06
Change np.int16 to np.uint32 in the response file reading

GallegoSav:
Pascal noticed using np.int16 could lead to negative value in the response matrices if the bincontent become to large. So I change to uint32 in order to be safe


israelmcmc:
Thanks @GallegoSav and @scipascal . Given the nature of the change --simple, contained, and with no new code--- I'll bypass the coverage requirements and merge this.

Issue #267, opened 2024-11-19
Extended Source Response Generation


hiyoneda:
I made schematics about how the classes related to the response matrix are connected through their class methods.
For the point source response, `FullDetectorResponse` is tied with `PointSourceResponse` by` FullDetectorResponse.get_point_source_response`, and the expected counts are calculated through `PointSourceResponse.get_expectation`.

When I implemented `ExtendedSourceResponse`, I was focusing on the expected count calculation part, i.e., `ExtendedSourceResponse.get_expection` and `ExtendedSourceResponse.get_expectation_from_astromodel`. So, currently, there is no function to generate `ExtendedSourceResponse` from `FullDetectorResponse`.

We can discuss where we should add such a method in cosipy, but I think it is natural to add a new method in `FullDetectorResponse`.
It can be based on some notebooks on which we generated the pre-computed response file for DC2 (while I don't know where the latest one is (and who has it)).

Regarding this, I have two questions.
- `SpacecraftFile` also has a method to generate `PointSourceResponse`; `SpacecraftFile.get_psr_rsp`. Do you think we should keep it or remove it in the future? 
- Do you think that we should rename `ExtendedSourceResponse.get_expection` to something like `ExtendedSourceResponse.get_expection_from_hist` and `ExtendedSourceResponse.get_expectation_from_astromodel` to `ExtendedSourceResponse.get_expection`? It is because the input of `PointSourceResponse.get_expectation` is astromodel function, but that of ``ExtendedSourceResponse.get_expection` is the model cube histogram. The naming is the same, but the input parameter is different, which may be confusing?

![extended_response_generation 001](https://github.com/user-attachments/assets/6c79c886-f979-4719-b73f-81dde09e0516)

Issue #129, opened 2024-01-25
Pre-computed all-sky point source response class (AllSkyPointSourceResponse)


israelmcmc:
This issue consists of creating a new class AllSkyPointSourceResponse which can calculate and handle the pre-compute all-sky point source response in inertial coordinates --e.g. galactic. It involves the code below in the comments (to create it), and moving [get_psr_in_galactic](https://github.com/Yong2Sheng/cosipy_SpacecraftFile/blob/f77143da464636947fdceac8f93eda5eb7aa8e87/cosipy/ts_map/fast_ts_fit.py#L146) to obtain the (single-pixel) PointSourceResponse class.

=========
Related comment from @ckarwin on slack https://ucbssl.slack.com/archives/C02TWC7U4NL/p1706213731961159?thread_ts=1700492800.163339&cid=C02TWC7U4NL

> Hi Israel, I'm converting the Ti44 response to the point source response using the same code we used for 511 (and Al26). I was running into errors until I remembered that I needed to add "return psr" before converting it to a tuple, as per your comment above. This was lost when I updated my cosipy version. Can you make this an option eventually? Maybe something like: if make_tuple==False: return psr? Or maybe you have in mind a better way to implement this? It's not urgent. I modified the code for now. But I wanted to mention this while it's fresh in my mind. I can also open an issue for the record.

israelmcmc:
The code is:
```
import time
from mhealpy import HealpixBase
import numpy as np

new_frame = 'galactic'

gal_coords = HealpixBase(nside = response.nside, coordsys = new_frame).pix2skycoord(np.arange(response.npix))

start = np.arange(0, gal_coords.size, 10)
stop = np.append(start[1:], gal_coords.size)

for pi,pf in zip(start, stop):

    print(pi,pf)
    
    t0 = time.time()

    coords = gal_coords[pi:pf]

    psr = response.get_point_source_response(coord = coords, scatt_map = scatt_map)
    
    psr.write(f"psr_gal_DC2/hist_{pi:03d}-{pf:03d}.h5")
    
    print(f"Pixels {pi:03d}-{pf:03d}: {time.time() - t0}")
```

Currently, we need to fudge the code temporarily and add `return psr`  before this line:
https://github.com/israelmcmc/cosipy/blob/0610c21f0567f9517d01ba89237a813258a2656e/cosipy/response/FullDetectorResponse.py#L897

This needs to be cleaned up and added to cosipy.

israelmcmc:
[...]

PR #284, opened 2025-01-23
Extended Source Response Generator

hiyoneda:
Regarding #129 and #267, I added new functions into FullDetectorResponse to generate an extended source response from a full detector response.

- get_extended_source_response : it generates an extended source response using a full detector response and cosipy.spacecraftfile.SpacecraftFile.
- get_point_source_response_per_image_pixel : it generates a point source response for a specific HEALPix pixel using a full detector response and cosipy.spacecraftfile.SpacecraftFile.
- merge_psr_to_extended_source_response : it creates an extended source response by merging multiple point source responses.
- _setup_extended_source_response_params : it is a supplementary function to setup NSIDE parameters for an image and a scatt map.

Basically, an extended source response can be generated like this.
```
from cosipy.spacecraftfile import SpacecraftFile
from cosipy.response import FullDetectorResponse, ExtendedSourceResponse

full_detector_response = FullDetectorResponse.open("SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.h5")

orientation = SpacecraftFile.parse_from_file("20280301_3_month_with_orbital_info.ori")

extended_source_response = full_detector_response.get_extended_source_response(orientation,
                                                                               coordsys='galactic',
                                                                               nside_scatt_map=None,
                                                                               Earth_occ=True)

extended_source_response.write("extended_source_response_continuum.h5", overwrite = True)
```
As default, the NSIDE of the scatt map is the same as that of the full detector response, but when we want to use a finer binned scatt map, its NSIDE parameter can be modified by setting the input parameter`nside_scatt_map`.

By using `get_point_source_response_per_image_pixel` and `merge_psr_to_extended_source_response,` we can generate a response on each pixel and merge the generated point source responses into an extended source response later. It can be used to generate a large extended source response using multiple nodes in a computer cluster.
On each node,
```
from cosipy.spacecraftfile import SpacecraftFile
from cosipy.response import FullDetectorResponse, ExtendedSourceResponse

full_detector_response = FullDetectorResponse.open("SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.h5")

orientation = SpacecraftFile.parse_from_file("20280301_3_month_with_orbital_info.ori")

ipix_image_list = [int(_) for _ in sys.argv[1:]]

basename = "psr/psr_"

for ipix_image in ipix_image_list:

    psr = full_detector_response.get_point_source_response_per_image_pixel(ipix_image, orientation, 
                                                                           coordsys='galactic',
                                                                           nside_image=None,
                                                                           nside_scatt_map=None,
                                                                           Earth_occ=True)
    
    psr.write(f"{basename}{ipix_image:08}.h5",overwrite = True)

```
When we merge them, the code can be like this.
```
from cosipy.spacecraftfile import SpacecraftFile
from cosipy.response import FullDetectorResponse, ExtendedSourceResponse

full_detector_response = FullDetectorResponse.open("SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.h5")

basename = "psr/psr_"

extended_source_response = full_detector_response.merge_psr_to_extended_source_response(basename, coordsys = 'galactic', nside_image = None)

extended_source_response.write("extended_source_response_continuum_merged.h5", overwrite = True)
```


hiyoneda:
I've compared the generated extended source response with that used in DC2. I used "SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.h5" as a full detector response, and "20280301_3_month_with_orbital_info.ori" as an orientation file. Then, I compared the response generated with the new function with "psr_gal_continuum_DC2.h5". Here, the earth occultation is not considered.

They are almost the same, while there is a slight difference of about 0.1% when I projected them onto the Ei axis. I suppose the slight difference may be caused by a different parameter setup, maybe the NSIDE of the scatt map.

download
download-1

When I generated the response by considering the earth occultation, the effective area gets smaller in a higher energy band. I think that it is a correct behavior because a higher energy gamma ray can penetrate the shield and then the effective area gets larger if the earth occultation is ignored.

download-2

Also, I confirmed that the response is identical with the standard method when I used `get_point_source_response_per_image_pixel` + `merge_psr_to_extended_source_response`.

hiyoneda:
With `nside_scatt_map=full_detector_response.nside*2`, the generated response becomes identical to `psr_gal_continuum_DC2.h5`. So, we can understand that the slight difference above was caused by the different `nside_scatt_map`.

ckarwin:
Thanks for testing that @hiyoneda. I'm going to test your code by producing the 511 PSR for DC3 on the NASA cluster. I should be able to do this early next week. If everything works ok then I can merge your PR.

hiyoneda:
@israelmcmc Can I ask to make a full detector response in a dense matrix for the unit test? Currently, the full detector response for the unit test is in the old format; its axes include 'SigmaTau' and 'Dist', and the response matrix is sparse. Actually, with it,  `get_point_source_response`, which is used for the extended response generation, raises an error because Coord +  scatt_map is currently only supported for dense responses.

ckarwin:
@hiyoneda I have been testing your PR and it works well. Very nice job! I have some suggestions before merging:

1. Please add unit tests. I guess your waiting for the detector response from @israelmcmc. 
2. The examples that you provided above work well and are very helpful. I think you should add this to the examples directory. It might not work so well as a Jupyter notebook, unless you use the same response from the unit tests. Or maybe you can just add python scripts. What do you thin  @israelmcmc, is there a good place in the docs directory for example scripts? Or is it better to do Jupyter notebooks?
3. Your example for `get_point_source_response_per_image_pixel` needs `import sys`. 

Currently, I'm calculating the 511 keV PSR for DC3 on the NASA cluster. If it looks ok, I think we can merge this PR after addressing the above comments.

israelmcmc:
Sorry for the delay in getting the test response. I'll work on it to not delay further the unit test.

Yes, I think it's time to start adding some scripts. The scripts and Jupyter notebooks serve a different purpose though, and they are not mutually exclusive. Jupyter notebook are meant to explain to the user how to use the library code, include more verbose than a script, and don't need to be very efficient. In general, we could have both a script and a Jupyter doing similar things.

In this particular case, I think this should be a script for production, not an example --i.e. it doesn't require the extra work to hold the users' hand and doesn't need to be in the docs folders.

We'll have a discussion next Thursday (Feb 6) about scripts. It was prompted by the SDOC group, but it also applies to this thread.

I think the way to go is to create scripts using [setuptools' entry points](https://setuptools.pypa.io/en/latest/userguide/entry_point.html). I wrote an example a while ago. See the entry_points item in [setup.py](https://github.com/cositools/cosipy/blob/develop/setup.py) and the corresponding [function](https://github.com/cositools/cosipy/blob/aea580b58514ee4e63dbb1837f360e3c2fe50d0f/cosipy/response/FullDetectorResponse.py#L1002). I think it's out of date now, but enough to get an idea of how to proceed.

augustus-thomas:
As a note on creating scripts that serve similar purposes to Jupyter Notebooks. If you want to save time, I have found that to convert a notebook to script quickly you can use `jupyter nbconvert --to script "path/to/notebook.ipynb"` . As @israelmcmc mentioned though, scripts should be more efficient and not exactly a copy-paste. Also this would not follow the entry-points guidance. But it could save development time and I wasn't sure if people were aware of it.

israelmcmc:
Thanks, @augustus-thomas! I was not aware of it. It does seem useful as a starting point for a script. I'm tagging @ldigesu who's working on converting some of our jupyter nb to executables in order the build the automatic pipelines. Please see above, Laura.

israelmcmc:
@hirokiyoneda Sorry for the delay. You can find the test file in  PR https://github.com/cositools/cosipy/pull/287
 
I changed the tests to use this file instead. I also committed a small change to the code that allowed me to create this file in an easier way.

hiyoneda:
@israelmcmc Thanks for making the new test response file. I've checked it and PR #287; they look great. I want to ask others who worked on the full detector response before; from my side, I think it can be merged once `open_rsp` is covered by the unit test. For this PR, I’ll make the unit test using the new dense test response file.

israelmcmc:
All sounds good, thanks @hiyoneda. I'll come back to the open_rsp test, although it's not a high priority at the moment.

ckarwin:
@hiyoneda I tested your PR by generating the 511 keV response on the NASA cluster. Everything worked fine, and I can generate the PSR very quickly (currently using 1000 parallel cores). As a sanity check, below is an example image taking a slice and projecting onto PsiChi:
NuLambda: 1440 (Galactic center)
Em: 0
Phi: 5

I believe that we still need to update COSILike in order to seamlessly use the precomputed PSR for analyzing point sources. I didn't see an issue for this, but it's documented in the code comments. I don't think it needs to be done for this PR, but maybe we can at least open an issue (I didn't see one). Please let me know what you think.  

Once you finish adding the unit tests I think this PR can be merged. 

Screenshot 2025-02-03 at 1 13 58 PM

israelmcmc:
Pretty cool, thank you both!

I'll open the issue about pre-computed PSR for point sources.

israelmcmc:
Here it is https://github.com/cositools/cosipy/issues/288

hiyoneda:
@ckarwin Thanks for testing the code with the NASA cluster. I am happy to see the attached figure. I have added the unit tests with the new test response which @israelmcmc generated. I have also added the example python scripts in `docs/tutorials/response/` (hope that I added them in the correct directory). I think that we've done things mostly.

ckarwin:
Looks good, and all checks have passed. @israelmcmc Are you ok with me merging this? Maybe we can implement the scripts as entry points later on?

israelmcmc:
Yeah, that's all right. We can move the scripts around later. Thanks!

ckarwin (approved):
All looks good now. Merging.

PR #287, opened 2025-02-03
Change unit test response file to dense

israelmcmc:
We have moved from a sparse to a dense response. It was not that sparse, and it ended up being slower than a dense matrix. The unit test now reflects what we actually use.

I created the dense response with:
```
from cosipy.response import FullDetectorResponse
from cosipy import test_data
from histpy import Histogram

dr = FullDetectorResponse.open(test_data.path/'test_full_detector_response.h5')

dr_new = Histogram.concatenate(dr.axes['NuLambda'], list(dr))
dr_new = dr_new.project(['Ei', 'NuLambda', 'Em', 'Phi', 'PsiChi'])
dr_new = dr_new.todense()

FullDetectorResponse._write_h5(dr_new, test_data.path/'test_full_detector_response_dense.h5')
```

In order to make things easier I moved the code that writes an HDF5 response file from a Histogram from _open_rsp to its own function, _write_h5. I added a unit test for this specific function.


israelmcmc:
I need to create a fake .rsp to fix the test coverage. @augustus-thomas I think at some point you were working on this, did you get to create one? Or maybe it was for the PSR, I don't remember exactly.

tests/response/test_full_detector_response.py
```diff
@@ -5,35 +5,61 @@
...
-    
+def test_write_h5(tmp_path):
+    """
+    Tests storing a Histogram as an HDF5 response
+    """
+
+    tmp_dir = tmp_path / "tmp"
+    tmp_dir.mkdir()
+
+    tmp_rsp = test_data.path / 'tmp_rsp.h5'
```
hiyoneda, line 36: maybe, should it be `tmp_path.path / 'tmp_rsp.h5'`? Currently, the file will be saved in the test data directory.

israelmcmc: Thanks for the catch. Indeed it was the wrong path. I fixed it in my latest commit.

ckarwin:
@israelmcmc I will try to check this tomorrow. I think it will help give me better coverage for my PR #300. Do 100% of new lines really need to be covered by unit tests for a new PR to pass the coverage requirement?

israelmcmc:
Thanks @ckarwin. Yeah, I think it's related to the covered code in #300. 

And nope, there's a 100% target with, currently, a 50% leeway ;) 
https://github.com/cositools/cosipy/blob/develop/codecov.yml

ckarwin:
Ah ok. I think the problem is that my current coverage is 0%, even though it's just four lines. After merging your PR I think it will be above 50%.

israelmcmc:
@ckarwin Yeah, good catch. I synced it with the latest version of v0.3.x and resolved a bunch of conflicts, including the one you mentioned (and resolved it to keep your changes). 

I also updated the PR to be merged with v0.3.x. I'll merge it with develop afterwards.

ckarwin:
Thanks for making the changes. I tested your latest code by opening one of the DC3 response rsp files, and it works fine. 

One more thing: I see the unit test now uses the dense response. Do you want to also repeat the test for the sparse response? Or do you prefer not to since it will require additional time? I was thinking that it would help the code coverage exceed 50%. Otherwise, we can override this requirement.

israelmcmc:
Cool! Thanks for checking.

I was considering removing the sparse response altogether. It seems like we'll no longer use it, and it's just one more piece of code we have to maintain. What do you think?

As for the code coverage, the main piece of code that is not being exercised by the test suite is opening an .rsp file and converting it to HDF5. The test I modified opens the HDF5. _write_h5 was a colateral, since I needed it to create the test HDF5 response.

ckarwin:
> I was considering removing the sparse response altogether. It seems like we'll no longer use it, and it's just one more piece of code we have to maintain. What do you think?

Hmm, maybe eventually, but I would leave it for now. I think it doesn't hurt to keep, and it could be useful at some point.

> As for the code coverage, the main piece of code that is not being exercised by the test suite is opening an .rsp file and converting it to HDF5. 

Ah yes, that's right. We should create a dummy rsp test response at some point. I used this method a lot for converting the DC3 response files from Andreas to h5. 

I'm ok to merge this now if you're ready.

ckarwin:
@israelmcmc Did you want me to hold off on merging this?

israelmcmc:
@ckarwin Yeah, I don't have a strong opinion, but since this is not needed for DC3, I thought it would be better to wait a little to see if I can add tests to cover my new code.

ckarwin:
I see. No problem. Yeah let's wait.

ckarwin:
This PR is needed for PR#314 (@Yong2Sheng , @GallegoSav), and so we decided to go ahead and merge.

ckarwin (approved):
Fine to be merged.

ckarwin:
@israelmcmc I approved this but I guess I can't push to this branch.

Issue #286, opened 2025-01-24
Handling the SAA passages 


GallegoSav:
For DC3 the SAA passages are simulated, so for sources and background we need to remove events when timestamp correspond to a SAA passage.

I was thinking that @ckarwin could do this cut during the conversion of the tra file into fits file. I share a example below using pandas which is quiet convenient for this type of tasks.

```
import pandas as pd

#we read the SAA LC
SAAProtonLC = pd.read_csv("/lustre/project/nhr-cosi/MC/DC3/INPUT/LightCurve/SAAproton_4MeVto2000.dat",skiprows=1,engine="python",skipfooter=2,names=["DP","time","flux"],sep=" ")
SAAProtonLC.set_index("time",inplace=True)


#your background or source file you read into a dataframe
df = pd.DataFrame(....)
df.set_index("timestamp",inplace=True)


#now we reindex the SAALC file to the dataframe time
SAAProtonLC_tmp = SAAProtonLC.reindex(df.index,method="nearest")
df["SAAflux"] = SAAProtonLC_tmp["flux"]

#we remove all the events where timestamp correspond to SAA flux >0
df = df[df["SAAflux"]==0]

```

Of course the exposure time will need to be corrected (for 3 months you remove ~16 days) but you can compute this once so we could create one function that compute the correct exposure time  for a specific time interval. Here a example 

```
df = pd.read_csv("/localscratch/sgallego/linkToXauron/MC/Background/COSIMEX/DC3/INPUT/LightCurves/SAAproton_4MeVto2000.dat",skiprows=1,engine="python",skipinitialspace=True,skipfooter=2,names=["DP","time","flux"],sep=" ")

# time period you want to study, I commented so this case will be the 3 months
#df = df[  (df["time"]<=TimeMin)  & (df["time"]<=TimeMax)    ]


df['datetime']= pd.to_datetime(df["time"],unit="s")
df.set_index("datetime",inplace=True)

# Calculate time deltas for each row
time_deltas = df.index.to_series().diff().fillna(pd.Timedelta(seconds=0))

#print(time_deltas)

# Identify periods where SAA_flux > 0 and where SAA_flux == 0
exposure_active = time_deltas[df['flux'] > 0].sum()  # Total exposure time for SAA_flux > 0
exposure_inactive = time_deltas[df['flux'] == 0].sum()  # Total exposure time for SAA_flux == 0

# Convert the Timedelta result to seconds, minutes, or hours as needed
[...]

PR #289, opened 2025-02-04
SAA Passage

ckarwin:
I modified the code to account for SAA passage. The live time corresponding to the SAA passage is now passed in the ori file, as shown below:
image (5)

I added methods in the dataIO class to cut photons detected during the SAA passage. The method first scans the ori file to find the bad time intervals, and then filters the good data. A sanity check is shown below. This is for DC3 simulations of a 511 keV point source at the Galactic center. In this case the SAA cuts a bit less than half the events.  
Screenshot 2025-02-04 at 1 09 07 PM

 
Screenshot 2025-02-04 at 1 10 15 PM

For the model fitting, this is implemented in cosilike via the scattmap calculation. As a sanity check I compared the MEGAlib simulations to the predicted counts from cosipy, shown below. They are in very good agreement. There is a discrepancy of 1.35%, compared to 0.43% before implementing the SAA cut. At this time it's not clear to me why the discrepancy has increased slightly, however, it is overall very small. I think it's ok to move forward with this for now and see if it can be refined in the future. It may just be due to the binning. 
Screenshot 2025-02-04 at 1 16 32 PM


GallegoSav (approved):
good to go for me :+1:

ckarwin:
Thanks, @GallegoSav. Do you want to merge?

avalluvan:
Hi Chris @ckarwin, 
Why does the SAA cut result in ~50% of the photons being cut. Based on issue #286 and my own observations, shouldn't this value be closer to ~17%?

ckarwin:
Hi @avalluvan That's a good question. I believe the ~50% cut is due to a bug that was found in the ori file, coupled with the fact that for the example described in this PR I used a point source located at the Galactic center. 

For more background information on the bug found in the ori file, please see the thread on the COSI slack here: https://ucbssl.slack.com/archives/C02TWC7U4NL/p1748014848318889

Essentially, for the DC3 orientation file, Earth rotation was not properly taken into account. For this reason, the orientation of COSI always points to the same place on the sky (in Galactic coordinates) during the SAA passage, which happens to be towards the Galactic center. Consequently, we are cutting a disproportional amount of events from sources located towards the Galactic center. 

You can see this in the "SAA Cut Sanity Check" plot above. If you look at the photon events w/o SAA cut, they always coincide with the SAA passage. Moreover, the SAA interval cuts about half of the distribution. 

This issue has been fixed regarding the orientation file. We decided to not make any changes for DC3. I will discuss this more at an upcoming cosipy meeting, but likely we will use the same simulations for DC4, and so updated simulations with this fix likely won't come till DC5.  

> Based on issue https://github.com/cositools/cosipy/issues/286 and my own observations, shouldn't this value be closer to ~17%?

Please let me know if this makes sense now, of if you still have some questions.

avalluvan:
Thank you for clarifying. Looks like this problem already has sufficient eyes on it.

PR #294, opened 2025-02-26
Update conf.py

ckarwin:
The API in cosipy docs is missing dataIO, Polarization, and Utilities, meaning it didn't successfully compile in the Spinx build. Checking the logs in the cosipy Actions tab, I see that the failure seems to be from the missing package: iminuit. So I added this to autodoc_mock_imports, and hopefully that fixes things.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -68,7 +68,8 @@ autodoc_mock_imports = ["histpy",
                         'scipy',
                         'psutil',
                         'awscli',
-                        'yayc'
+                        'yayc',
+                        'iminuit'
                         ]
 
 # There seems to be a conflict between unittest.mock (used by sphinx) and metaclasses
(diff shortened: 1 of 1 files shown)


ckarwin:
Well, that doesn't seem to work...

ckarwin:
Let me try another way.

israelmcmc:
@ckarwin You could also try with MOCK_MODULES, in the same file. For some reason I don't understand, some modules don't work with the autodoc's version of mock modules.

ckarwin:
@israelmcmc Ok, I can try that, but now I'm having difficulty compiling the docs even if I remove the new import. I'm not sure if there is another issue. Are you able to compile ok?

israelmcmc:
There was more than one issue. Please see #296

israelmcmc:
@ckarwin ^^^

PR #296, opened 2025-02-26
Fix docs

israelmcmc:
There are 2 issues preventing the docs from building:
- Missing iminuit in mocked modules
- The polarization modules had an operation "-np.inf" in the default value for one of the functions signature. Since the dependencies modules are mocked, they can't have operation in the signature. The default is instead initialized in the function body.

I also added the line bkg estimation example to the tutorial, since I noticed it was missing.


israelmcmc:
🤔 It worked in my machine, but there still seem to be some errors. I'll keep poking at it.

ckarwin:
Thanks! Yes, it's strange.

israelmcmc:
@ckarwin Indeed, it seems to be an issue with nbsphinx
https://github.com/sphinx-doc/sphinx/issues/13352

Apparently, sphinx fixed an issue in their newer version, but nbsphinx was working correctly thanks to the issue sphinx fixed! Hehe.

I merged your PR into this PR, you'll see your commits in the commits tab. I think we still need to merge this one since I fixed other issues in addition to your changes. Please take a look.

ckarwin (approved):
Looks good.

ckarwin:
@israelmcmc I guess this won't actually show up in the public docs unless we merge this into main, right? Should we do that now?

israelmcmc:
Thanks. Yeah, I just merged it with v0.3.x and main. It's geeting deployed: https://github.com/cositools/cosipy/actions/runs/13573407083

It's messy because main is supposed to be reserved for releases. 

I've been working on having multiple version within the same documentation page, like ReadTheDocs, so we don't have to do this, but I haven't finished.

PR #297, opened 2025-02-26
Update the orientation file and response file in the SpacecraftFile and TS map tutorial notebook

Yong2Sheng:
I updated the tutorial notebooks for SpacecraftFile and TS map:
- Use orientation file orbital information
- Use response file with earth occultation


Yong2Sheng:
OK, I just realized an issues caused by the response with earthocc: it causes infinities during the TS calculation. So we should revert back to the original response.

ckarwin:
@Yong2Sheng 
> OK, I just realized an issues caused by the response with earthocc: it causes infinities during the TS calculation. 

Can you be more explicit? What exactly is happening?

>  So we should revert back to the original response.

What do you mean? Earth occultation needs to be included.

Yong2Sheng:
Hi @ckarwin, I think it's because the the current dwell_time map calculation doesn't consider the earth occultation, so we have the case:
- expectation = 0
- data != 0.

Using `scatt_map`(it considers the earth occultation) should avoid this issue. I plan to try it out since the `scatt_map` will replace the `dwell_time` anyway, it a good time to update it for DC3.

ckarwin:
@Yong2Sheng 
Well what we do in COSILike to add a small number to the expectation:
`expectation += 1e-12 # to avoid -infinite log-likelihood (occurs when expected counts = 0 but data != 0)`
https://github.com/cositools/cosipy/blob/d63897c1a1fa85991dfd29dee995ccde355e4c1d/cosipy/threeml/COSILike.py#L290C9-L290C113

You could try something similar if needed. 

So is your PR ready to be checked, or do you want to change it to a draft for now?

Yong2Sheng:
Hi @ckarwin, I would like to have it drafted for now. I will implement what you said, and then merge the PR. Thanks!

ckarwin:
Ok, please go ahead and make it a draft, and then reopen when it's ready to be reviewed. Thank you.

israelmcmc:
I have the feeling this was already solved by another PR, wasn't it @Yong2Sheng? Can we close it?

Yong2Sheng:
Hi @israelmcmc, it' true for the TS map. I think we can close it and then I can make a new PR to update the SpacecraftFile tutorial notebook. (it will be just changing a few lines in the SpacecraftFile notebook to update the orientation file.). Does this sound OK for you?

israelmcmc:
Thanks @Yong2Sheng. Sounds good. I'm closing it for now then.

PR #298, opened 2025-02-27
Update __init__.py

ckarwin:
Polarization class is not being initialized in top level init script. I noticed this b/c the polarization docs have a different format than the other docs.

### DIFF
--- a/cosipy/__init__.py
+++ b/cosipy/__init__.py
@@ -18,3 +18,7 @@ from .source_injector import SourceInjector
 
 from .background_estimation import LineBackgroundEstimation
 from .background_estimation import ContinuumEstimation
+
+from .polarization import conventions
+from .polarization import polarization_angle
+from .polarization import polarization_asad


israelmcmc:
Hmmm this one I'm not sure. The `cosipy/__init__.py` exposes classes that we want to import directly as e.g.:
```
from cosipy import PolarizationASAD
```
instead of 
```
from cosipy.polarization import PolarizationASAD
```
To do that, you'd need to add to `cosipy/__init__.py`:
```
from .polarization import PolarizationASAD
```

You are intead importing the files in `polarization` as submodules, which I don't think is what we want or what the rest of the modules do. Instead, this file decides which classes from the file become part of the polarization module:
https://github.com/cositools/cosipy/blob/develop/cosipy/polarization/__init__.py

What is the issue you are seeing in the docs? Maybe there's another way to solve it.

ckarwin:
Hmm, ok, I was understanding wrong then. For many of the other classes (for example dataIO), the file name is the same as the class name. I thought `cosipy/__init__.py` was referring to the file, but I guess it's the class. That makes sense. 

For the polarization docs, it might not be a big issue, but I'm not sure what the convention is. Most of the other modules show the main classes. For example, dataIO shows the classes: DataIO, UnBinnedData, BinnedData, and ReadTraTest. The polarization docs just shows "Conventions". I'm not sure how this module is structured. 

This is probably just a coding convention that can be refined later. Please let me know.

ckarwin:
Looking here: https://cositools.github.io/cosipy/api/index.html

israelmcmc:
I got what you meant now. I think I fixed it. Can you please take a look at that link again?

The issue was that the polarization documentation was the only page that contained subsections. The classes were hidden under one of these subsections, which didn't appear in the index because their nested level was too high. I increased the depth and added a "Classes" subsection for all modules.

ckarwin:
Hmm, now I see 3 different bullets for polarization conventions, but no classes. Is that how it's supposed to look?

ckarwin:
Nevermind, I see now. Yeah looks good. I'll close this PR. Thanks!

israelmcmc:
Cool! No problem 👍

ckarwin:
@israelmcmc fixed issue.

Issue #293, opened 2025-02-25
Error opening rsp response file with polarization


ckarwin:
I'm trying to open the DC3 polarization response:
```
from cosipy.response import FullDetectorResponse
rsp_file = ResponseContinuum.o3.pol.e200_10000.b4.p12.s10396905069491.m441.filtered.nonsparse.binnedpolarization.11D.rsp.gz
FullDetectorResponse.open(rsp_file)
```

I get the following error: 
```
Traceback (most recent call last):
  File "/gpfsm/dnb34/ckarwin/COSI/Data_Challenge/DC3/Response/convert_response.py", line 16, in <module>
    response = FullDetectorResponse.open(each)
  File "/gpfsm/dnb34/ckarwin/Software/COSIPY/cosipy/cosipy/response/FullDetectorResponse.py", line 88, in open
    return cls._open_rsp(filename,Spectrumfile,norm ,single_pixel,alpha,emin,emax)
  File "/gpfsm/dnb34/ckarwin/Software/COSIPY/cosipy/cosipy/response/FullDetectorResponse.py", line 630, in _open_rsp
    data=np.transpose(dr_area.contents, axes = [1,0,2,3,4]),
  File "/discover/nobackup/ckarwin/Software/COSIPY/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 655, in transpose
    return _wrapfunc(a, 'transpose', axes)
  File "/discover/nobackup/ckarwin/Software/COSIPY/lib/python3.10/site-packages/numpy/core/fromnumeric.py", line 59, in _wrapfunc
    return bound(*args, **kwds)
ValueError: axes don't match array
```

Tracing this back, the error stems from line 630: https://github.com/cositools/cosipy/blob/f25cc32fb676bf89299ef888ad0a378f1412552a/cosipy/response/FullDetectorResponse.py#L630

This is for non-sparse matrices, and you see there are only 5 axes included. However, a bit higher in the code you can see that for polarization response 6 axes are defined: 
https://github.com/cositools/cosipy/blob/f25cc32fb676bf89299ef888ad0a378f1412552a/cosipy/response/FullDetectorResponse.py#L544

**So I think the fix to this is to modify near line 630**, and add an if statement for response files with polarization.` If polarization, then axes = [1,0,2,3,4,5]; if no polarization, then axes = [1,0,2,3,4].` I tested this with the response file above and it seems to work. 

[...]

PR #300, opened 2025-02-27
Update FullDetectorResponse.py

ckarwin:
This fixes issue #293. I tested it with the DC3 polarization response to verify that it works.

### DIFF
--- a/cosipy/response/FullDetectorResponse.py
+++ b/cosipy/response/FullDetectorResponse.py
@@ -624,18 +624,15 @@ class FullDetectorResponse(HealpixBase):
 
         #non sparse
         else :
-           
-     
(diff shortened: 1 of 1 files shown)


ckarwin:
Spacing is fixed now. All checks passed.

israelmcmc:
I'll bypass the patch coverage requirements since none of the lines were covered anyway. It's unrelated to this change. See #287 .

Since we're preparing for a release, I'll change the target to v0.3.x, and then merge to develop. That is easier than doing the reverse.

israelmcmc:
Done. Merged with develop as well. 

Thanks @ckarwin

ckarwin:
Awesome. Thanks.

PR #302, opened 2025-03-10
Update track_overflow bin

saurabhmittal23:
- Load full detector response on memory without the underflow/overflow bins to prevent memory leak


israelmcmc:
@jdbuhler Similar to #255, do you think #307 and #343 superseds this PR as well?

PR #304, opened 2025-03-11
Image Deconvolution tutorial cleanup for DC3 release

ckierans:
I've worked through Hiroki's image deconvolution notebooks which haven't been updated since DC2. Modified some of the language to make it more general. Updated the ori file to be used which is compatible with recent updates in DC3.

PR #309, opened 2025-03-21
Image Deconvolution tutorial cleanup for DC3 release (rebased)

israelmcmc:
Copying from #304:

> I've worked through Hiroki's image deconvolution notebooks which haven't been updated since DC2. Modified some of the language to make it more general. Updated the ori file to be used which is compatible with recent updates in DC3.

I rebased it to remove duplicated commits with different hash


ckierans:
Thanks, Israel. Not sure how i managed the rebase/different commit hashes but I appreciate you cleaning it up. @hiyoneda  this should be good to merge now if you agree with the updated language.

PR #313, opened 2025-03-25
file modified to create pipeline task cosi_threemlfit

ldigesu:
Hello @israelmcmc. I added the test for the code in tests/pipeline. This is just running the task and checks if the outputs exist in cosipy/test_data. No additional test data were needed in this case. The test works ok for me locally, but when loaded here apparentely it fails, maybe users are not allowed to save outputs?  apart from that, there are other tests that are failing, not related no modifications I made. Let me know If you have feedback about this. Thanks and cheers,


ldigesu:
The check that is failing is due to missing coverage. In order to pass it you need to write tests that execute the new lines of code you introduced with this commit (the code in cosipy/pipelines). There's some help about the: https://github.com/cositools/cosipy/blob/develop/docs/dev/unit_tests.rst . You can also see multiple other examples from other modules.

>>>>>>  I am not sure I understood this. My understanding is that I need to write code in the test directory that executes the task.py. This will imply adding the test dataset that I used. 

The new script is calling cosipy.pipeline.task.task:cosi_threemlfit, which doesn't exist. I think you committed the file .idea/.name instead of task.py. I suppose you are using pycharm. If that's the case, it might work better if you open a pycharm project at the cosipy root directory. You still need to delete the .idea/.name file, it belongs to your IDE, not the repo.

>>>>>>>>I think I have now added all the files that are needed. In any case, I don't know how to remove a file that was committed.


Tks,

israelmcmc:
@ldigesu 
> I am not sure I understood this. My understanding is that I need to write code in the test directory that executes the task.py. This will imply adding the test dataset that I used.

Yes, this is correct. Typically we add a light version of the dataset you used, which while they won't provide scientific-quality results, they are good enough to test the code itself. There are already various files in the cosi.test_data directory. Since you are not adding new functionality to the core library, but a script to run existing functions, I don't expect you to need new files. Please check the existing tests for the spectral fitting code, I suppose you need something very very similar: https://github.com/cositools/cosipy/blob/develop/tests/threeml/test_spectral_fitting.py

> I think I have now added all the files that are needed. In any case, I don't know how to remove a file that was committed.

Use `git rm` https://git-scm.com/docs/git-rm

israelmcmc:
@ldigesu I'm back at reviewing this. For now, I resolved the conflict in #353. Please review those changes. I'll continue reviewing the rest in the meantime.

ldigesu:
> @ldigesu I'm back at reviewing this. For now, I resolved the conflict in #353. Please review those changes. I'll continue reviewing the rest in the meantime.

reviewed and merged #353 (LD)

israelmcmc:
@ldigesu I'm sorry we couldn't finish the dicussion on the config file and parameter format yesterday. I still want to provide a concrete example of how all of this would look like. I modified your branch to match my proposed format from the slides I shared yesterday. You can see the changes in #355. We'll keep discussing this next week, but please take a look in the meanwhile and please let me know if there are any concerns.

In addition to that, I also noticed a few other things in your PR:
- I'm not sure if the time cuts are working. At least I couldn't test it with the files you have in the unit tests, since that data is not binned in time. Do you have an example using the time cuts?
- While the unit tests you provided do check that your code runs without raising exceptions, I think they should also check that the outputs have the expected results. This can be done with `assert` statements. You can see other unit tests for examples.
- With my proposed changes to the config file and parameters, I think some of the function in the pipeline module are no longer needed. Specially the 3ML parsing, since I'm using the one provided by astromodels. Maybe there are other function that are also not needed, or that are so small that it might not be worth to have them as individual function. Please take a look.

ldigesu:
Hello, @israelmcmc, I merged #355 in my branch, made small modifications to have the time cuts working with Time objects as input, and added the possibility to append a suffix to the names of the output files. The override option does work correctly. I also tested the option of giving tstart, tstop, and output_dir via terminal, and everything works fine. Let me know if you want me to share with you the files that I used to test the time cuts. Let me know if you need anything else. All my changes are committed here. Thanks and cheers,

israelmcmc:
Thanks, Laura. Looks good to me. I'm merging.

PR #303, opened 2025-03-10
Bug fix for source interval method

ckarwin:
This PR fixes issue #299. When I added the livetime to the spacecraft file, I didn't update the source interval method, which essentially makes a sub-selection of the ori file. That has now been updated. I tested it by running the following: 

```
from cosipy.spacecraftfile import SpacecraftFile
from astropy.time import Time

ori_file = "DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori"
ori = SpacecraftFile.parse_from_file(ori_file)
new_ori = ori.source_interval(Time(1842123855.0221176, format = 'unix'), Time(1842123887.7090132, format = 'unix'))

print("original ori start:")
print(ori._time[0].value)
print(ori._time[1].value)
print(ori.livetime[0])

print("new ori start")
print(new_ori._time[0].value)
print(new_ori._time[1].value)
print(new_ori.livetime[0])

print("original ori end:")
print(ori._time[-1].value)
print(ori._time[-2].value)
print(ori.livetime[-1])

print("new ori end")
print(new_ori._time[-1].value)
print(new_ori._time[-2].value)
print(new_ori.livetime[-1])
```

Which gives this:

```
original ori start:
1835487300.0
1835487315.0
15.0


new ori start
1842123855.0221176
1842123870.0
14.977882385253906


original ori end:
1843467255.0
1843467240.0
0.0


new ori end
1842123887.7090132
1842123885.0
2.7090132236480713
```

So it seems to work as expected, but please test for yourself.

I also made a small change to the UnBinned class. For the function that parses tra files, I added a check to the photon event data, ensuring that all the needed info is present in the tra file. Currently, it only checks that each photon event has at least two hits in the detector. If not, the event is not included in the photon list. This is an issue that came up when parsing the primary proton backgrounds for DC3. It only occured for 4 out of millions of events, and so it's very rare.  

@eneights This addressed the issue that you found a couple of weeks ago. If you can test this PR it would be much appreciated!

PR #315, opened 2025-03-25
Bug fix for source interval method (rebased hotfix)

israelmcmc:
Apply (rebased) commits from #303 that should have been in branch v0.3.x but were merged onto develop instead.


israelmcmc:
@ckarwin double-checked this. I'm merging.

PR #318, opened 2025-03-31
propose fix for issue 317

zh3nl:
Proposed fix for issue 317:

1. Introduced a default value of None for target_coord, essentially making the argument optional
2. Include a conditional check within the function that checks whether earth_occ is True. If so, a target_coord value must be provided, or else a value error will be raised.

### DIFF
--- a/cosipy/spacecraftfile/SpacecraftFile.py
+++ b/cosipy/spacecraftfile/SpacecraftFile.py
@@ -541,7 +541,7 @@ class SpacecraftFile():
         return self.dwell_map
 
     def get_scatt_map(self,
-                       target_coord,
+                       target_coord=None,
                        nside,
                        scheme = 'ring',
                        coordsys = 'galactic',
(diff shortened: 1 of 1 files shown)


israelmcmc:
Thank you @zh3nl! Your changes so far look good to me. One more thing please: it's good practice to list the positional arguments before the optional arguments --i.e. args before kwargs. Can you please swap the order of `nside` and `target_coord`, and update the usages of `get_scatt_map` you see in the rest of cosipy accordingly? Using an IDE like pycharm might be the easiest way.

The unit tests are also failing due to the method signature change. They should be fixed once you implement the changes I suggested above.

zh3nl:
Hey @israelmcmc, I swapped the order of positional arguments to be before the optional arguments (i.e., nside is moved before target_coord). For your second statement, do you suggest I manually change the unit tests? Or do you mean that the unit tests will be fixed once I implement the changes?

israelmcmc:
Thanks @zh3nl! 

>  For your second statement, do you suggest I manually change the unit tests? Or do you mean that the unit tests will be fixed once I implement the changes?

Yeah, it seems that the unit tests need to be updated as well. Just the ones that explicitly call get_scatt_map though.

Also, I think that you missed this one:
https://github.com/zh3nl/cosipy/blob/issue-317/cosipy/threeml/COSILike.py#L345-L346

zh3nl:
Sounds good, I will close the issue now. Have a great rest of your week.

israelmcmc:
Closed! Have a great week as well!

PR #323, opened 2025-04-02
PR 318 Follow up: optional target_coord when earth_occ = None

israelmcmc:
#318 Made target_coord optional when earth_occ = False. However, the code was still using target_coord even if earth_occ = False, which produced an error since the target_coord default is None. I moved all calls to target_coord to the blocks where earth_occ = True

@zh3nl Can you please take a look and let me know what you think? I'd like to have a second pair of eyes before merging this.

### DIFF
--- a/cosipy/spacecraftfile/SpacecraftFile.py
+++ b/cosipy/spacecraftfile/SpacecraftFile.py
@@ -600,19 +600,20 @@ class SpacecraftFile():
         # Get max angle based on altitude:
         max_angle = np.pi - np.arcsin(r_earth/(r_earth + altitude))
         max_angle *= (180/np.pi) # angles in degree
-        
-        # Calculate angle between source direction and Earth zenith
-        # for each time stamp:
-        src_angle = target_coord.separation(earth_zenith)
-        
-        # Get pointings that are occulted by Earth:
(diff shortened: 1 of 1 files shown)


zh3nl:
I will take a look at this when I can and get back to you on it.

zh3nl:
Looks good so far. One thing I do want to ask is, are there any actions/functionalities that still run when earth_occ is False and target_coord is optional? If so, are they needed? Also, are there calls to the method (within the same files or in other files) where target_coord was passed in/used when earth_occ is False?

israelmcmc:
Thanks @zh3nl!

> are there any actions/functionalities that still run when earth_occ is False and target_coord is optional? If so, are they needed? 

Not that I know of. target_coord is used if and only if earth_occ=True

> are there calls to the method (within the same files or in other files) where target_coord was passed in/used when earth_occ is False?

Maybe there are, I'm not sure. However, if earth_occ=False and a target_coord was passed, it will just be ignored --which was also true even before target_coord was made optional.

zh3nl:
Then I think it should be good? All the unit tests seem to pass and we can always make another issue when it arises.

israelmcmc:
Thanks @zh3nl! I'll merge it then.

PR #328, opened 2025-04-03
Fix path for continuum background estimation tutorial

israelmcmc:
- Update ori file to the one with the new format with orbital information and uptime
- Fix response path. .h5.zip -> .zip

### DIFF
--- a/docs/tutorials/background_estimation/continuum_estimation/BG_estimation_example.ipynb
+++ b/docs/tutorials/background_estimation/continuum_estimation/BG_estimation_example.ipynb
@@ -285,7 +285,7 @@
    "metadata": {},
    "source": [
     "The notebook requires the following files:\n",
-    "1) DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth.ori\n",
+    "1) DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori\n",
     "2) SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.earthocc.h5\n",
     "3) crab_bkg_binned_data_galactic.hdf5\n",
     "4) inputs_crab.yaml\n",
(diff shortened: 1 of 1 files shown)


ckarwin:
@israelmcmc Thanks for catching this! I looked at the updated notebook, and found a few things that also need updating:

The cell after the imports cell:

```
The notebook requires the following files:

DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth.ori
SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.earthocc.h5
crab_bkg_binned_data_galactic.hdf5
inputs_crab.yaml
They can be downloaded using the cells below.
```
Can you update the ori file and response file names here too?

The cell where the detector response is defined seems to still have the wrong ending:
```
data_path = "your/path"

# Orientatin file:
ori_file = os.path.join(".","DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori")

# Spacecraft orientation:
sc_orientation = SpacecraftFile.parse_from_file(ori_file)

# Detector response:
dr = os.path.join(data_path,\
        "SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.nonsparse_nside8.area.good_chunks_unzip.earthocc.h5")

crab = SkyCoord(l=184.56*u.deg,b=-5.78*u.deg,frame="galactic")
psr = instance.calc_psr(sc_orientation, dr, crab)
```
I think the detector response file shouldn't have the `.h5`

Have you tried running the download cells to make sure they are working? No need to fully download, just make sure it starts ok. If you could confirm that they are working ok that would be much appreciated. I'm on the NASA system and it actually won't let me connect to wasabi (I have everything locally).

israelmcmc:
@ckarwin Thanks for checking this. I updated the .ori file in the cell after the upload. However, the response does have the suffix .h5 after unzipping it, despite the zipped file being .zip and not .h5.zip.

ckarwin:
Ah, ok, I see. Thanks for checking.

ckarwin (approved):
Only small updates, and they are ok.

israelmcmc:
Thanks!

PR #330, opened 2025-04-04
Update source_injector to use new get_scatt_map signature

israelmcmc:
The get_scatt_map signature recently changes in #318. This only affected develop and not v0.3.x since I didn't want to change the API of the released version. While #318 updated the existing calls to the method, this particular instance was only recently added to v0.3.x. When I merged v0.3.x into develop to keep it in sync, this made the tests fail. This should fix it.


israelmcmc:
Thanks @Yong2Sheng!

PR #301, opened 2025-02-28
Polarization fitting in galactic coordinates

eneights:
I updated the point source response code to handle polarization, and updated the ASAD method code to be able to perform the fit in both spacecraft and galactic coordinates. 
The fitting in galactic coordinates does not yet work properly, and I will still need to add unit tests and update the notebook.
This should resolve issue #293. 
Also closes issue #210

PR #332, opened 2025-04-06
Polarization fitting in galactic coordinates (PR301 rebased and follow up)

israelmcmc:
These are @eneights commits from #301, but rebased onto v0.3.x instead of develop. 

In addition to solving conflicts that arose from the rebase, there are 2 other small change:
- The plots are not shown by default during the tests. Otherwise the user had to close the windows when run interactively.
- I found a small bug that caused the extended source response tutorial to crash. There was an if condition that was meant to be "X and (Y or Z)", but was instead "X and Y or Z", which was interpreted as "(X and Y) or Z".

I checked that all tutorials run properly.


israelmcmc:
@eneights Can you please double checked that all your changes are here, and confirm that you are OK with my extra changes? After that this can be merged.

israelmcmc:
@eneights When you are done with this one, can you please also take a look at #333. It has only one extra commit.

eneights (approved):
All of my updates appear to be here, and your changes look good!

israelmcmc:
Thanks! I'm merging it

PR #333, opened 2025-04-06
Sync develop with v0.3.0 (Polarization fitting in galactic coordinates)

israelmcmc:
Merge after #332 

This contains all commits from #332, and in addition, it fixes an issue arising from a method signature change in #318 (which is only part of develop, not v0.3.x)


eneights (approved):
Looks good to me!

israelmcmc:
Thanks! I'm merging it

PR #334, opened 2025-04-06
Automatically run all tutorials

israelmcmc:
The goal of this PR is to have a way to automatically run all tutorials and ensure they are still working, especially for PRs involving the modification of multiple files in the library.  

The main file of this PR is `/docs/tutorials/run_tutorials.py`. It can be run like this:

```
./run_tutorials.py run_tutorials.yml -o /path/to/output/dir --wasabi_mirror /path/to/wasabi/mirror/dir
```

This will automatically 
1. Fetch all needed files from wasabi and store them in the Wasabi mirror path
2. Execute all Jupyter notebook tutorials.

The `/docs/tutorials/run_tutorials.yml` configuration file defines the tutorials to be run. It lists the ipynb of the tutorial and all the files it needs to run.

You can also specify specific tutorials with e.g. `--tutorials test dataIO` which will run e.g. the "test" tutorial (a dummy one I added for testing purposes) and the "dataIO" tutorial.

The tutorials are run in a clean directory in the `-o` path. After they finish, you'll see two files ending in `_executed.ipynb` and `_executed.html`, which are the result of running the full nb. 

Once all tutorials are run, the script prints a summary like this, specifying whether the notebook executed successfully --i.e. if they got to the end, the user still needs to check if the results make sense-- and how long they took (in seconds)
```
2025-04-06 13:05:51 - INFO - Run summary:
background_continuum:
  elapsed_sec: 184.3989166659885
  succeeded: true
background_line:
  elapsed_sec: 375.44333629199537
  succeeded: true
dataIO:
  elapsed_sec: 236.0974838329712
  succeeded: true
extended_source:
  elapsed_sec: 1330.1044526250334
  succeeded: true
imaging_511_scatt:
  elapsed_sec: 813.4683292909758
  succeeded: true
polarization_asad:
  elapsed_sec: 229.7361044170102
  succeeded: true
response:
  elapsed_sec: 78.74918104201788
  succeeded: true
source_injector:
  elapsed_sec: 95.59728316700784
  succeeded: true
spacecraft_file:
  elapsed_sec: 33.47875029197894
  succeeded: true
spectral_fit_crab:
  elapsed_sec: 105.73264166602166
  succeeded: true
spectral_fit_grb:
  elapsed_sec: 53.7675430419622
  succeeded: true
test:
  elapsed_sec: 6.997152458992787
  succeeded: true
ts_map:
  elapsed_sec: 1321.9503156249993
  succeeded: true
```

I also modified:
- The notebooks, so they download the wasabi files and can be executed "as is", without the user having to modify the paths.
- `fetch_wasabi_file`, so it skips if the file already exists and has the correct checksum. It also can unzip files.

Longer term, I envision:
- Having this script run automatically somewhere (a cluster?) daily so we can catch quickly when we break the tutorials
- Remove all .ipynb files in the docs folder that can't be checked automatically with this script -i.e. if you want to keep your .ipynb file, modify it so it runs automatically and add it to run_tutorials.yml.
- Require submitters and reviewers to run this for large PRs.

NOTE: This PR adds boto3 to the list of requirements (in setup.py). You can get it by re-installing cosipy, so with ` pip install boto3`

NOTE: run_tutorials.py requires [nbconvert](https://nbconvert.readthedocs.io/en/latest/install.html). It can be installed with `pip install nbconvert`. I didn't want to add it to the list of cosipy dependencies in setup.py since it's not really needed by the library itself.


israelmcmc:
Closing and reopening to try to update commit list

israelmcmc:
NOTE: This PR adds boto3 to the list of requirements (in setup.py). You can get it by re-installing cosipy, so with ` pip install boto3`

NOTE: run_tutorials.py requires [nbconvert](https://nbconvert.readthedocs.io/en/latest/install.html). It can be installed with `pip install nbconvert`. I didn't want to add it to the list of cosipy dependencies in setup.py since it's not really needed by the library itself.

ckarwin:
@israelmcmc I'm running this and overall it's working well. Great work! I do have 1 issue:

I have my wasabi_mirror directory in place. When I rerun the code I see that in most cases it detects that the needed wasabi files are already present and skips the download. However, for the files that need to be unzipped it doesn't seem to be skipping the files that are already there. Specifically, my run keeps getting killed on the NASA cluster (presumably b/c it is taking lots of RAM to unzip) when it comes to this line:
`INFO - Fetching and unzipping COSI-SMEX/DC2/Responses/PointSourceReponse/psr_gal_511_DC2.h5.gz`

But I already have the unzipped file in my wasabi_mirror, and so I think the code should detect this before trying to unzip the file, and move on. Is there way to implement this?

ckarwin:
Another issue it that I tested running it for a single tutorial but I don't see where the summary file is printed. I see a log file in the corresponding tutorial directory, but not a run summary like you posted above. Is this supposed to be saved somewhere?

ckarwin:
@israelmcmc 

> Can you please update your copy of my branch an try again?

I verified that I'm using your most recent version. 

> If that doesn't work, next thing is to make sure your file has the same checksum. You can check that with md5sum COSI-SMEX/DC2/Responses/PointSourceReponse/psr_gal_511_DC2.h5. It should be daaace1b814728b8410aee0367e449a3

I tried this and it works fine, and I have the same checksum. For context, I'm running on the login node of the Discover cluster. A job on this node is automatically killed if it starts using too much RAM. 

I think I tracked down the issue. It comes from this command in the util function: 
`local_checksum = md5(open(unzip_output, 'rb').read()).hexdigest()`
While the other method of getting the checksum works fine, running this command from the python terminal also kills the run. As a further test, I commented out this line in the source code, and replaced it with the correct checksum for the file above, and the code runs with no problems.

I think the issue with the command above is that the file is being opened, which requires lots of RAM. Is there way to instead get the checksum using the alternative way that you posted above? This seems to require less RAM.  

Thanks for considering my other suggestions. I'll test them when they are implemented.

israelmcmc:
Thanks for debugging it. Yeah, it seems that I need to chunk it up, just like DataIO!
https://stackoverflow.com/questions/1131220/get-the-md5-hash-of-big-files-in-python

I'll work on it

israelmcmc:
@ckarwin Please take a look again:

- I synced my branch with develop
- All files now have a checksum. This should prevent querying wasabi at all (assuming I got all the needed files right)
- There's an option ` --verify-wasabi-mirror` to check that you have all necessary files in your mirror with the right checksum
- The MD5 sum is faster and uses less memory for big files
- The summary is now printed to run.log in the output dir.
- I changed the format of the summary to add green/red color.

israelmcmc:
The tests  are filing due to a problem with codedov, it seems. I'll take a look later.

ckarwin:
Thanks for making these changes, @israelmcmc. I'll test it now.

ckarwin:
@israelmcmc Your recent changes fixed all the issues I mentioned above. Thanks again! I was able to successfully run all the tutorials in parallel on the cluster. It took less than an hour, with the max time corresponding to the notebook that takes the longest to execute. It's very cool to be able to check all the notebooks so efficiently! 

As shown below, all of the notebooks succeed except for the response notebook. I guess this was initially passing based on your summary above. Maybe things changed when you synced your branch with develop? Can you please double check? For reference, my run log for the response tutorial is attached. 

```
SUCCEEDED: test
SUCCEEDED: spacecraft_file
SUCCEEDED: spectral_fit_grb
FAILED: response
SUCCEEDED: source_injector
SUCCEEDED: spectral_fit_crab
SUCCEEDED: dataIO
SUCCEEDED: polarization_asad
SUCCEEDED: background_continuum
SUCCEEDED: ts_map
SUCCEEDED: background_line
SUCCEEDED: imaging_511_scatt
SUCCEEDED: extended_source
```
Note: I wrote my own little script to parse the log file that is written in the Run directory and print this summary, since I'm running the tutorials in parallel.  

After resolving this issue, I think your PR can be merged. 

[response_run.log](https://github.com/user-attachments/files/19688919/response_run.log)

ckarwin:
It seems it was some sort of glitch. After rerunning the response tutorial it's working fine, both from the terminal and when submitting as a batch job.

israelmcmtest:
Thanks @ckarwin!

PR #335, opened 2025-04-07
Update polarization tutorial to use RelativeX response

eneights:
I modified the ASAD method notebook to download and use the RelativeX response, as well as do the spacecraft frame fit in the RelativeX convention


eneights:
Messed things up, starting over

PR #336, opened 2025-04-07
Update polarization tutorial to use RelativeX response

eneights:
Modify ASAD method notebook to download and use RelativeX response, as well as do spacecraft frame fit in RelativeX convention

### DIFF
--- a/docs/tutorials/polarization/ASAD_method.ipynb
+++ b/docs/tutorials/polarization/ASAD_method.ipynb
@@ -85,7 +85,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


israelmcmc:
Thanks @eneights! I think your PR #335 to v0.3.x is enough. For #332/#333 I made two PRs because the merge to develop involved a function with different arguments in v0.3.x vs develop, but I think your changes apply equally well for both branches. After merging #335 I'll merge v0.3.x into develop, that way we'll keep the same commit in both. I'll close this PR.

israelmcmc:
I guess #337 now, instead of #335

PR #337, opened 2025-04-07
Update tutorial to use RelativeX response

eneights:
Modify ASAD method notebook to download and use RelativeX response, as well as do spacecraft frame fit in RelativeX convention


israelmcmc:
Looks good to me! I'm merging. This is how they look for me for reference (using #334)

RelativeZ

RelativeX

PR #339, opened 2025-04-09
PR307 and v0.3.x merge conflict resolution. Follow up on "Initial patches to make cosipy work with histpy 2.x"

israelmcmc:
This is PR #307 merged with v0.3.x. 

I resolved some conflicts, mostly trivial, with the exception of what's in commit bd86d927a376f25913fae8569479386f9e835b5c. PR #307 was fine when it was first created, but PR #301 introduced some code that was incompatible without some small changes.


jdbuhler (approved):
Looks OK to me.  This doesn't involve my code changes directly, though I see it's touching a lot of the binary test files (which my changes also touch to remove overflow padding).

Issue #249, opened 2024-10-08
Issue when plotting raw_lightcurve


NicoloParmiggiani:
Hi,

I found this issue when I try to plot the raw_lightcruve. By default the show_plots value is set to False. However in the documentation there is no mention to set it to True.

![Screenshot 2024-10-08 alle 18 20 53](https://github.com/user-attachments/assets/bad85f4f-5f31-47c7-b040-aa1cfa25923f)

@israelmcmc

israelmcmc:
Thanks @NicoloParmiggiani. 

Tagging @ckarwin for awareness.

ckarwin:
Thanks for pointing this out @NicoloParmiggiani. I'll fix this next week during COSIFest.

Issue #338, opened 2025-04-08
DataIO tutorial plots not showing


AdrienLaviron:
In the DataIO tutorial as run per COSI DC3 instructions the plots are not showing as expected in Jupyter. This is due to argument show_plots defaulting to False in most BinnedData functions. Simple solution is to add `show_plots=True` in the functions that are supposed to display plots.

israelmcmc:
Thanks @AdrienLaviron!

I can confirm. This is how it looks for me using #176: https://gist.github.com/israelmcmc/2377064fdb5d9add2de8145ed0b5a5e7

Short term
We should indeed add `show_plot = True` in the tutorial. But only in the tutorial, the default should still be `false`.

Longer term
I suggest the following API:
```
unbinned_data = UnbinnedData("/path/to/unbinned_data.tra.gz")
binned_data = unbinned_data.get_binned_data(axes)
binned_data.plot()
```

@ckarwin What do you think?

israelmcmc:
Closed by #342. See also #344.

PR #342, opened 2025-04-09
now showing plots in dataIO tutorial

ckarwin:
This PR is a short term solution to issue #338. The code was modified since we initial created the dataIO tutorial, and a `show_plots=True` argument must now be passed to some of the functions in order for the plots to be shown. I added this to the tutorial as needed. 

I think we can keep issue #338 open even after this PR is merged. The dataIO API can be refined in the longer term to make things more intuitive (see the discussion in #338 for more details).

### DIFF
--- a/docs/tutorials/DataIO/DataIO_example.ipynb
+++ b/docs/tutorials/DataIO/DataIO_example.ipynb
@@ -553,7 +553,7 @@
     }
    ],
    "source": [
-    "analysis.get_raw_spectrum()"
+    "analysis.get_raw_spectrum(show_plots=True)"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


ckarwin:
This also resolves issue #249. I think it would be better to close both #249 and #338 and then open a new issue about refactoring (low priority).

israelmcmc:
Thanks @ckarwin . This looks good to me, I'm merging. 

>This also resolves issue https://github.com/cositools/cosipy/issues/249. I think it would be better to close both https://github.com/cositools/cosipy/issues/249 and https://github.com/cositools/cosipy/issues/338 and then open a new issue about refactoring (low priority).
I agree, I'll go ahead an open it

israelmcmc:
See #344.

PR #345, opened 2025-04-11
Updated line_background_estimation_example_notebook.ipynb

hiyoneda:
I've updated line_background_estimation_example_notebook.ipynb to use the DC3 primary proton background file.


israelmcmc:
Thanks @hiyoneda! Just couple of things:

* I made a PR to your PR to update the file in the `run_tutorials.yml` file as well. This by itself doesn't affect the notebook, but it allows us to check that it works automatically as described in #334. Can you please add this to your branch?
* I executed the notebook, and while it finished and the plots look OK, I noticed that the migrad fit failed (or has a warning):

Screenshot 2025-04-11 at 11 05 17 AM

Have you seen this before? Any idea why I'm getting this but your commited version of the executed nb doesn't seem to have this problem?

For reference/convenience, in the `tutorials/background_line` inside this file: [tutorials.zip](https://github.com/user-attachments/files/19708482/tutorials.zip)
You'll see:
- line_background_estimation_example_notebook.ipynb (the version in this PR)
- line_background_estimation_example_notebook_executed.ipynb, which is my executed version
- line_background_estimation_example_notebook_diff.html, the diff between the version in this branch and my executed one
- All outputs

Btw, I got this using
```
/path/to/run_tutorials.py /path/to/run_tutorials.yml -o tutorials --wasabi_mirror /path/tp/wasabi/cosi-pipelines-public/ --tutorial background_line
```

As described in #334

hiyoneda:
I confirmed the same output as what you sent me. Previously, I used an old histpy version in which the overflow/underflow bins exist by default. Maybe this difference affected the results. 

Actually, now the lines are not fitted well, e.g., some Gaussian functions are fitting the same line in the bkg (and some lines are ignored), and a Gaussian function's normalization is negative after fitting. Considering this case, I want to add some functions, e.g., fixing some parameters and limiting a range of some parameters. Note that I am submiting a new PR. So for this PR, I think that it is ok to be closed.

israelmcmc:
Thanks for checking @hirokiyoneda! Yeah, the overflow can make a difference, even if there are no overflows in the energy itself. If the histogram is keeping track of all under/overflow, then a multi-dimensional histogram projected into a given axis is guaranteed to be the same as if you had binned the data only in the particular axis from the beginning. Otherwise, it can be different, as you could be losing events inadvertently from the other axes.

I do see this warning in the notebook:
```
fill() discarded one or more values due to out-of-bounds coordinate in a dimension without under/overflow tracking
```

This doesn't prevent the code from being merged, but it's worth checking into.

I'll close this PR in favor of #348

PR #352, opened 2025-04-22
Custom 3d GALPROP model

ckarwin:
I've added a custom 3d function to astromodels that takes an input GALPROP sky map in healpix format (i.e. a data cube). I modified cosipy accordingly to use this in COSILike. Specifically, I had to modify the ExtendedSourceResponse class in order to work with 3D functions. Unit tests have been added to provide coverage for most new additions. I also wrote an example tutorial notebook demonstrating how to use the custom GALPROP model. The tutorial has been setup to automatically run with the new machinery now in `cosipy/docs/tutorials`. It's been tested and runs ok. 

I'll note that astromodels actually already has a GALPROP function, but it's only compatible with GALPROP FITS format, and it's currently broke anyway. I plan to look into this more soon, and try to make a PR in astromodels with the fix. I'll probably eventually try to add my GALPROP model to astromodels directly as well, but for now I think it's a good start to just put it in cosipy custom functions, which is what I did.

This model applies specifically to GALPROP inputs, but more generally, a similar model could be developed to take any input data cube (in FITS or healpix format). I think this wouldn't require too much modification, and I will probably work on this in the near future.  

As part of the example tutorial, I fit galdiff + albedo. A comparison of the predicted source counts to the input source counts is shown below. I get agreement to within ~7% for all energy bins. 

Screenshot 2025-04-21 at 10 41 51 PM

Likewise, the total predicted counts matches very well with the total input counts, as shown below. The agreement is within ~1%.

Screenshot 2025-04-21 at 10 45 28 PM

The best-fit normalizations for the galdiff and background are:
galprop map: 0.7721 +/- 0.0008 
background: 1.00771 +/- 0.00029 
Thus, I recover the background to within <1%, although the galprop normalization is lower than the input by ~23%. The discrepancy is plausibly related to the course spatial binning of the response, which has an nside=8, corresponding to an area of 7.3 x 7.3 square degrees. The intensity is currently being calculated at each fixed position from the response, and then multiplied by the solid angle, likely leading to an overestimate (the map has a strong gradient). Possible remedies for this include either using a higher resolution response, or modifying the function to average over a solid angle corresponding to the resolution. I will try both of these methods in the next week. Note that the fit uses the precomputed extended source response, which accounts for Earth occultation and SAA passage. Below is the recovered average intensity (averaged over all-sky) compared to the input model.

Screenshot 2025-04-21 at 10 48 17 PM

For reference, the input sky map (calculated at 501 KeV) is shown below. It's the model used for DC3. 

Screenshot 2025-04-21 at 11 04 06 PM


ckarwin:
@hirokiyoneda If you have time, you might be a good person to look over this. A lot of what I did mirrors what you already setup with the ExtendedSourceResponse. I just needed to develop new methods to work with 3d models.

hiyoneda:
@ckarwin, yes, I will review this PR.

ckarwin:
Thanks, @hiyoneda. 

I have one question regarding your code that maybe you can help clarify for me. Considering the line below: https://github.com/cositools/cosipy/blob/2b9da1ae22674a96aa3386419e31ed0cac6202ca/cosipy/response/ExtendedSourceResponse.py#L104
My understanding is that you are calculating the `allsky_image_model` at coordinates corresponding to the response, which gives intensity at the specific points (i.e. flux/sr). You then multiply by the pixel area of the response to get the flux in each pixel. Is that correct? 

I am also doing the same thing. However, the normalization I recover for the GALPROP map is ~22% lower than the input simulations. I thought maybe this could be due to the coarse binning in psichi, but that doesn't seem to be the case.  I've tried a few different approaches where I integrate over the pixel area, but so far it doesn't seem to change things very much.

hiyoneda:
Yes, that is correct. The contents of the binned response have a unit of ph * cm2 * s. So when we convolve it with an model map, we need to integrate the map over each pixel, and currently I simply multiply the pixel size with it. 

I feel that we miss something in the fitting because 23% sounds huge. For example, the SAA is removed from the event list, but the SAA time intervals are included in the extended response generation. Do you think something like that could happen?

ckarwin:
@hiyoneda Thanks for confirming. I think it's a reasonable approach. 

Yes, I agree that the 23% difference is too big. Just to remind, this discrepancy pertains to the normalization of the input model, although the predicted counts is in pretty good agreement (within ~5%). From what I can tell, the SAA times should be removed from the extended response generation, but maybe it's somehow related to removal of the SAA events. For example, I tried fitting the simulated data without the SAA photons removed, and in this case the normalization is much better -- within ~5% of the model input. 

I think I'll try generating the extended response again from scratch to see if I get the same thing.

ckarwin:
I reproduced the extended source response from scratch, and confirmed that I still get the same result.

hiyoneda:
I have checked the code. It is well written! So far, I have not found any significant issue that can cause ~ a 20% difference in the fitted spectrum. I am speculating that there are some issues in the response, orientation, etc. (outside of the codes). I am going to check the response and orientation next week.

ckarwin:
Thanks, @hiyoneda! Ok, please keep me posted. I plan to try to do the fit with the 04 continuum response to see how it changes things. I'll try to get to this next week as well.

hiyoneda:
@ckarwin I am checking the orientation file used in your tutorial notebook, and noticed that all of the delta time values between adjacent data points in the orientation are 15 seconds. Does this mean the SAA time intervals are not removed in the orientation file? If the extended source response (extended_source_response_continuum_merged.h5) was generated using this orientation file, the exposure time may be overestimated due to unremoved SAA intervals. According to Savitri, the net exposure after the SAA cut is ~70 days, which may cause ~22% overestimation for the exposure and ~22% underestimation for the flux (very consistent with the reported discrepancy!).

Screenshot 2025-05-16 at 14 55 01

ckarwin:
@hiyoneda The SAA time intervals are not removed from the ori file. It includes all timestamps with a 15 second cadence. However, I calculated the extended source response using [get_point_source_response_per_image_pixel](https://github.com/cositools/cosipy/blob/24f31fb9d96627792acc36fc98edae95b2965479/cosipy/response/FullDetectorResponse.py#L974), which uses the scatt_map. The weights in the scatt_map come from the livetime in the ori file, which is 0 when in the SAA, and 15 otherwise. So I think the exposure time should be ok, but let me know if I'm missing something here. 

But somehow it seems like the issue may indeed be related to the SAA. When I run the analysis using the src file without the SAA photons removed, the normalization I get is close to 1.

hiyoneda:
I don't think that the SAA is removed in [get_point_source_response_per_image_pixel](https://github.com/cositools/cosipy/blob/24f31fb9d96627792acc36fc98edae95b2965479/cosipy/response/FullDetectorResponse.py#L974). In my understanding, weighting is performed for the Earth occultation effect. We need to add an additional file to specify when the SAA intervals are because they cannot be determined from the satellite attitude only.

hiyoneda:
Sorry, the SAA information is stored in the orientation file. The last column is the livetime for each point, and it is set to zero when the satellite is in the SAA. Is it correct? Perhaps, something unexpected happens about the livetime or weighting calculation.

ckarwin:
> Sorry, the SAA information is stored in the orientation file. The last column is the livetime for each point, and it is set to zero when the satellite is in the SAA. Is it correct?

Yes, exactly.

hiyoneda:
@ckarwin I was checking if [get_point_source_response_per_image_pixel](https://github.com/cositools/cosipy/blob/24f31fb9d96627792acc36fc98edae95b2965479/cosipy/response/FullDetectorResponse.py#L974) considers the SAA. At the position of Crab, the generated point source response has the same effective area before/after considering the SAA cut (I modified the last column in the orientation file so that all values in the column are 15 sec, to ignore the SAA cut). For the zenith, it differs depending on the SAA cut. So, it is very likely that `get_point_source_response_per_image_pixel` has some bugs. I will check it more tomorrow.

ckarwin:
@hiyoneda Thank you for the update. That sounds like a promising lead. Please keep me posted. I have also been thinking that the problem is likely somehow related to the SAA, but so far I haven't been able to figure it out.  I'll be very interested to see if you can find a bug in `get_point_source_response_per_image_pixel`.

As another test, I ran the analysis using the high-resolution continuum response: `SMEX/DC3/Data/Responses/ResponseContinuum.o4.e100_10000.b10log.s10396905069491.m166.filtered.nonsparse.binnedimaging.imagingresponse_nside16.area.good_chunks.h5.zip`
To do this, I first generated the precomputed point source response. I get very similar results as the initial analysis. Specifically, the normalization of the GALPROP component is still low by ~22%. I think this indicates that the issue is not due to a the response being too coarse. I also checked the simulated source file, and verified that the SAA photons are properly removed. For reference, these results are shown below. 
galdiff_test

hiyoneda:
I plot the z-pointing in the galactic coordinates when the livetime is zero in the orientation file. It seems that the satellite points to a particular region in the galactic coordinates. Do you think it is correct? I felt that the SAA passing time would shift a bit due to the Earth's rotation, but I don't know...

```
z_pointing_map = np.zeros(hp.nside2npix(32))

for z_pointing in (ori.z_pointings[:-1])[ori.livetime == 0]:
    z_pointing_map[hp.ang2pix(32, z_pointing.l.deg, z_pointing.b.deg, lonlat = True)] += 1.0

hp.mollview(z_pointing_map)
hp.graticule(dpar = 10.0, color = 'white', alpha = 0.5)
```

download

ckarwin:
@hiyoneda I think the plot is at least consistent with what I've seen before. Essentially, the z-axis pointing traces two different trajectories on the sky, for the north-south pointings. Maybe there is actually a shift in your plot. Does a single SAA passage encompass the entire band? Or does the length of the band result from slight shifts from the different passages?

hiyoneda:
> Does a single SAA passage encompass the entire band? Or does the length of the band result from slight shifts from the different passages?

I will check it very soon.

If the z-pointing map is correct, the fact that the exposure for Crab is not affected by the SAA seems OK. The location of Crab is at the edge of the map, and it is separated far away from all of the z-pointings in the SAA passage. The contribution to the effective area from these internals would be quite small or zero because of the small effective area and the Earth occultation.
Do you know that the number of events in the Crab event file does not change after the SAA cut?

ckarwin:
The number of events for the Crab doesn't change with the SAA cut.

hiyoneda:
The SAA passage always looks the same.

Since the number of events for the Crab doesn't change in the event file either, this is OK for this issue, but I feel a bit strange. The SAA location is fixed on the Earth, i.e, close to Brazil. Imagine that we are standing in Brazil and looking up at the sky every 90-100 minutes. I think that stars should move, and we see different parts of the sky for each SAA passage.

download-1
download-2
download-3
download-4

hiyoneda:
@ckarwin Today, I've confirmed that the input Galprop file (GalTotal_SA100_F98_input.dat) is consistent with the model calculation with GalpropHealpixModel and total_healpix_57_SA100_F98_example.gz. Now it looks like everything is correct...

Would it be possible to perform the spectral fitting by fixing the background normalization to one? It is because the galactic diffuse is much fainter than the background (Albedo in the notebook's case). So the normalization difference of ~1% in the background could cause ~10% normalization change on the source. I want to see how the issue will change by removing this factor.

ckarwin:
@hiyoneda Thanks for checking the input GALPROP file. That was a good test.


> Would it be possible to perform the spectral fitting by fixing the background normalization to one? 

Instead of fixing the BG to 1, what I did was fix the source normalization to 1. I think this addresses your question, but let me know if not. You can see the results below:

Screenshot 2025-05-23 at 9 54 51 AM

Screenshot 2025-05-23 at 9 55 36 AM

You see the largest discrepancy comes from bins 1, 2, and 10.

ckarwin:
@hiyoneda I am starting to think that the discrepancy might actually be coming from the coarse energy binning. It is somewhat implied in the shape of the fractional difference that I posted above. We see something similar with the DC2 crab tutorial, where the normalization is lower by ~8%. The Galactic diffuse is even worse (~22%), but it's not clear how this impacts an all-sky source compared to a point source. The crab case is also a bit more convoluted because there are other parameters involved. 

To test this I plan to simulate the same GALPROP model using a flat spectrum (which is what the response uses). This can be accomplished by making the MEGAlib input dat file using the same energy slice for all energy bins.

ckarwin:
> Since the number of events for the Crab doesn't change in the event file either, this is OK for this issue, but I feel a bit strange. The SAA location is fixed on the Earth, i.e, close to Brazil. Imagine that we are standing in Brazil and looking up at the sky every 90-100 minutes. I think that stars should move, and we see different parts of the sky for each SAA passage.

That's a good point. Let's ask @saurabhmittal23.

ckarwin:
@hiyoneda I ran the test with the flat energy spectrum. I started with the same GALPROP model, and then created the flat spectrum using the same energy slice for all energy bins. Specifically, I used energy bin 10, corresponding to the skymap at 1 MeV. As I suspected, I now get very good agreement for both the counts and source normalization, as summarized in the figure below. 

I think this explains the discrepancy of the GALPROP model normalization. Hopefully this issue will be resolved with the new response parametrization and handling, being worked on by @israelmcmc et al. But since the issue does not seem to be a problem with this PR, I think it can be merged now, if you agree. Please let me know. 
flat_spectrum_results

ckarwin:
@hiyoneda Here is another related example from a while ago. In this case I was trying to understand the discrepancy in the crab spectral fit for DC3. Again, fitting with a flat spectrum resolved the issue, indicating that the problem arises from the coarse energy bins of the response. I can try a couple of intermediate values of spectral index. 
crab_flat

israelmcmc:
This look good to me. @hirokiyoneda should approve it since he was already reviewing it. Based on today's meeting it seems a couple of tests with intermediate spectral indices would be enough to close the loop on this.

ckarwin:
@hiyoneda @israelmcmc I ran the test as we discussed. I created a GALPROP model with a much harder index, although not flat, i.e. intermediate between the original spectrum and the flat spectrum. Indeed, the results are as I expected. The best-fit norm in this case is 0.91. So the GALPROP normalization for the three cases gives:
```
original spectrum: 0.7721 +/- 0.0008
harder spectrum: 0.907 +/- 0.005
flat spectrum: 0.970 +/- 0.001
```

A summary of the latest test with the harder spectrum is given below. Please let me know if you are convinced by this result, and if the PR can be merged now. Thanks. 

temp1

temp2

israelmcmc:
Thanks @ckarwin. Looks good to me.

hiyoneda (approved):
@ckarwin Thanks so much. Now it is reasonable to me. Thank you for being patient with me. I approved it.
@israelmcmc @ckarwin Do you think it is worth opening an issue about the fact that the normalization depends on the spectral index? Maybe it is better to keep it somewhere.

ckarwin:
Thanks, @hiyoneda. No problem. It's good to be careful with these things. Do you mind merging the PR as well, since you approved it? 

I think it's not a bad idea to open an issue to track the problem. Maybe you can link this PR as well, as specific example.

israelmcmc:
I agree with opening an issue. Thanks. Please tag it with the label "coarse_response". I want to re-evaluate at some point if all of these are really due to the response (by either having a finer response or the unbinned analysis)

hiyoneda:
i will open an issue for this later.

PR #355, opened 2025-04-30
PR313 follow-up: proposed config and arguments format

israelmcmc:
Same functionality as #313, but I'm changing the config file and argparse arguments with a proposed general format.

### DIFF
--- a/cosipy/pipeline/src/fitting.py
+++ b/cosipy/pipeline/src/fitting.py
@@ -49,7 +49,7 @@ def build_spectrum(model, pars, par_minvalues, par_maxvalues):
 
 
 
-def get_fit_results(sou,bk,resp_path,ori,l,b,souname,bkname,spectrum):
+def get_fit_results(sou,bk,resp_path,ori,bkname,model):
     """
     Fits the spectrum to the data
     """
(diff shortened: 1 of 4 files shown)


ldigesu:
Hello Israel, I started testing this. Let me know how we proceed. Will you implement the comments and then merge in my branch, or the other way around?

In any case, for now, I noticed the following:
i) The override option does not work for me. Likely, it's me doing something wrong. E.G.The command cosi-threemlfit --config ./my_config --override “sc_file:./sc.ori” results in this error:      config.override(*args.override)
  File "/Users/lauradigesu/opt/anaconda3/lib/python3.9/site-packages/yayc/configurator.py", line 124, in override
    key,value = args[0].split("=")
ValueError: not enough values to unpack (expected 2, got 1)

ii) I tested the time cuts, and they work for me. However, all the functions that are called there need numerical values as input, not a Time object. Thus, I made everything work by adding tstart.value/tstop.value anywhere in lines 103-108 of task.py. My working example is grb of 88s of dc3, however I'll think about how an example for this using the internal test_data of cosipy.


iii) maybe it would be useful to add the option of adding a user-defined prefix or suffix to the names of the outputs? Since everything does not get overwritten by default, maybe one wants to keep track of which source is being fitted directly in the names of the output files.

Cheers,

ldigesu (commented):
Hello Israel, I started testing this. Let me know how we proceed. Will you implement the comments and then merge in my branch, or the other way around?

In any case, for now, I noticed the following:
i) The override option does not work for me. Likely, it's me doing something wrong. E.G.The command cosi-threemlfit --config ./my_config --override “sc_file:./sc.ori” results in this error: config.override(*args.override)
File "/Users/lauradigesu/opt/anaconda3/lib/python3.9/site-packages/yayc/configurator.py", line 124, in override
key,value = args[0].split("=")
ValueError: not enough values to unpack (expected 2, got 1)

ii) I tested the time cuts, and they work for me. However, all the functions that are called there need numerical values as input, not a Time object. Thus, I made everything work by adding tstart.value/tstop.value anywhere in lines 103-108 of task.py. My working example is grb of 88s of dc3, however I'll think about how an example for this using the internal test_data of cosipy.

iii) maybe it would be useful to add the option of adding a user-defined prefix or suffix to the names of the outputs? Since everything does not get overwritten by default, maybe one wants to keep track of which source is being fitted directly in the names of the output files.

iv) I agree that the build_spectrum function does not serve anymore

Cheers,

israelmcmc:
@ldigesu Thanks for checking this. I worked on top of your branch, so you can merge this onto your branch and keep commiting to it.

The override flags need an "=" sign. e.g. `--override “sc_file=sc.ori”. I'm not sure the `./` will work here, but give it a try.

 In general, I think we should work with Time objects since they are less error prone, giving that they keep track of the format. They can also account for leap seconds, and have 128bit precision. Differences in Time object give you a TimeDelta object, that you can also use.

I agree adding an optional flag for the output suffix would be useful.

PR #360, opened 2025-05-27
Adding PsiChi plot in the source injector

GallegoSav:
I added an option `make_PsiChi_plot` for getting the expected PsiChi map from the model.


ckarwin:
@GallegoSav What is the status of this PR? It seems to contain changes from the CO sequence, which should be part of PR #362. Also, I'm not sure why the entire unbinned data class has been recopied, and is actually duplicated in the file. Something is not right here. Also, the new code needs coverage. Please let me know the status.

PR #362, opened 2025-05-27
Adding the Compton Sequence variable during tra file reading

### LINKED ISSUE #361: Adding the Compton sequence in the list of variables for the CO events
Since the CO events with only 2 interactions have a lot of bad reconstructed events (i.e order of interaction is swap) , the number of interaction for the CO event , aka the compton sequence (SQ) in the tra file is an important  variable we would like to use for selection cut.  

The information is in the tra file, one would have to just modify the tra file reading part and add a "CO sequence" variable for the fits file. 

@ckarwin I can do it if you want.
@israelmcmc Should I need to wait for your big structure change or not ? That would be a DC4 milestone.

The example below show the PsiChi map for the crab. You can see a second compton cone due to misreconstructed event. On a sphere the ring would be on the other side of the crab position.  Asking for a cut CO seq >2 would most likely remove this second compton cone

Image

GallegoSav:
I only modified the unbinned part , assuming people will use the cut option on the unbinned file and then save it. 
I added a cut function similar to the energy one.


ckarwin:
@GallegoSav Is this reedy to be reviewed? If so, please add the necessary coverage and I will check the code.

cosipy/data_io/UnBinnedData.py
```diff
@@ -545,7 +557,7 @@ def write_unbinned_output(self, output_name):
 
         # Data units:
         units=['keV','s','rad','rad',
-                'rad','rad','rad','rad','cm','deg','deg']
+                'rad','rad','rad','rad','cm','deg','deg','']
```
avalluvan, line 566: I don't think any of the subsequent code directly uses the units written to the .fits/.hdf5 file. However, do you think it will be useful to include a more explicit string for the dimensionless quantity (number of compton interactions)? For example, 'number' or 'dimensionless'.

cosipy/data_io/UnBinnedData.py
```diff
@@ -718,6 +730,47 @@ def select_data_energy(self, emin, emax, output_name=None, unbinned_data=None):
 
         return
 
+    def select_data_COseq(self, seqmin, seqmax, output_name=None, unbinned_data=None):
+
+        """Applies CO sequence cuts to unbinnned data dictionary. 
+        
+        Parameters
+        ----------
+        seqmin :  int
```
avalluvan, line 745: Could you explicitly mention that the function returns the range [seqmin, seqmax), i.e., seqmin is included and seqmax is excluded?

avalluvan (commented):
I think the pull request is complete barring the two minor line comments. It will preserve the number of Compton interactions information which can later be used for event cuts and potentially be propagated to the BinnedData object.

cosipy/data_io/UnBinnedData.py
GallegoSav: I remember I had some issues if no units were declared.

GallegoSav:
> @GallegoSav Is this reedy to be reviewed? If so, please add the necessary coverage and I will check the code.

@ckarwin I added the coverage and updated the test scripts. The pull request is ready for your review

ckarwin:
@GallegoSav I am reviewing this PR. I tested the dataIO tutorial notebook and it runs ok. 

However, I did find one issue. Take a look at the block of code linked below, where we remove bad photon events: https://github.com/GallegoSav/cosipy/blob/11f64dfe1e00659111bc2dd223caced304c648de/cosipy/data_io/UnBinnedData.py#L206

This removes photons that only have one detector hit. It's pretty rare, but it does happen sometimes. I guess this would correspond to events with CO_seq=1. Since we append all the event parameters event by event before knowing if it's a good event or not, we have to remove the data afterwards for bad events. Essentially, I think you need to add `CO_seq.pop()` under `tt.pop()`.

Another thing to be aware of is that we are planning to move all event selections to an event selection class, utilizing the proper protocol, as part of the work from @israelmcmc. Also note that this event selection is a lower-level selection, and it will require the calculation of a new response, since the same selection needs to be made for the response. Maybe we can merge this PR for now, but then work on moving it to the new protocol in the near future. What do you think @israelmcmc? Do you prefer we start now working with the proper protocol? It's probably not a bad idea. If so, maybe we can just merge the part that saves the CO_seq number to the cosi_dataset, and move the selection method.

ckarwin:
@GallegoSav Can you also update your branch so it's synced with develop?

ckarwin:
Ok, thanks for making the change. Regarding the select function, @GallegoSav and I discussed it and decided to keep it in dataIO for now. We can make one new PR to move the select functions from dataIO to the new selection class. This PR can be merged now.

ckarwin (approved):
Can be merged.

Issue #199, opened 2024-07-16
Change precision of response in HDF5 file


israelmcmc:
We should use float16 instead of the default 64 to save memory and disc space.

israelmcmc:
Related to #110

israelmcmc:
Closed by #364

Issue #110, opened 2023-11-20
Adjust the chunk size of HDF5 response files


israelmcmc:
https://github.com/cositools/cosipy/blob/a14d9667ead0206e0dbad0acf27a2ccd1cb2e14d/cosipy/response/FullDetectorResponse.py#L631

See: https://davis.lbl.gov/Manuals/HDF5-1.8.7/Advanced/Chunking/index.html

Related to compression speed discussed in #106 

![Screenshot 2023-11-20 at 11 04 00 AM](https://github.com/cositools/cosipy/assets/34419239/bbec6261-165d-451e-8ca7-f7b700f29661)

israelmcmc:
Closed by #364

Issue #106, opened 2023-11-16
Reading zipped response files


ckarwin:
Is it possible for the response class to take h5.gz or h5.zip files?

israelmcmc:
Some context and thoughts I'd have:

The files in wasabi ending in .h5 are actually compressed by zip, just internally. HDF5 uncompressed them on the fly, but only the portion you are accessing. 

In general, it wouldn’t be great having to decompress the whole file every time you need to access some of it, as it would be the case if you accept h5.gz files. It is fine for small files, but for large files like this one, especially once we go to finer resolution, it might not be practical. 

I found that this on-the-fly decompression was slowing the rotation of the response significantly. That’s why I ended up creating an .h5 without internal compression that can be decompressed once and for all. 

I see this as a temporary workaround though. I think there might be a way to write the code to reduce the time spent on the internal decompression. I think right now it is decompressing the same data multiple times. I investigated this briefly, and part of the issue is that the HDF5 cache fills up before we access the same piece of data again.

I don't think that accepting whole zipped files is they way to go, as [internal compression is one of the main features of HDF5](https://www.hdfgroup.org/2017/05/hdf5-data-compression-demystified-2-performance-tuning/). Let's leave this issue open to remind us about all of this.

ckarwin:
Thanks for the detailed answer. Ok, sounds good.

israelmcmc:
Closed by #364

PR #364, opened 2025-05-28
Improved Full Detector Response storage and conversion

jdbuhler:
This PR overhauls the on-disk and in-memory format of the FullDetectorResponse. The principal goals are to

 * reduce on-disk storage size via HDF5 internal compression, without incurring an unacceptable running time penalty
 * reduce rsp->h5 conversion time and memory usage
 * allow h5->rsp conversion without data loss

The biggest change is that the raw CDS ring counts are stored as small integers on disk, separately from the effective area multipliers.  This change greatly improves compressibility and reduces peak memory during rsp->h5 conversion.  As a bonus, it allows more efficient weighted averaging of slices at runtime and dynamic selection of 32- vs 64-bit PSRs.

Conversion has been separated from the FullDetectorResponse class into its own RspConverter class, which handles both rsp->h5 and h5->rsp.

After consultation with Israel et al., these changes remove support for the following deprecated formats and features:
 * sparse responses
 * miniDC2 format
 * reading the spectrum for effective area computation from a file (was already broken in DC3 codebase)

The new on-disk response format is *not* backwards or forwards compatible with the previous format.  For this reason, I've established a new directory COSI-SMEX/develop in the public cosipy bucket on wasabi, put the new response files there, and updated the tutorials to use them instead of the ones from DC2/DC3.  It is anticipated that DC4 will use the new format, while DC3 will remain with the current one.

The patch introduces a dependency on the hdf5plugin package, a well-supported companion to h5py that implements the BitShuffle compression algorithm used by the new response format.

### Conversion time and memory

DC2 O3 continuum response:  < 4 minutes, 4 GB
DC3 O3 polarization response:  < 6 minutes, 6.33 GB
O4 version of continuum response: < 90 minutes, 52.7 GB

Testing was on a 2.3 GHz Intel Xeon Gold 5118 server with 192 GB of RAM and SSD storage; only one core was used.

(The first two used to take hours, while the last took overnight and needed at least half a terabyte of RAM)

### HDF5 response file sizes on disk

DC2 O3 continuum response: 592.76 MB  (.rsp.gz is 841.58 MB)
DC3 O3 polarization response: 795.76 MB (.rsp.gz is 954.58 MB )
O4 version of continuum response: 4187.05 MB (.rsp.gz is 5186.87 MB)

These file sizes are < 1/10 the size of the previous uncompressed HDF5 responses for O3, and more like 1/50 for O4
response, assuming (as was the case for the file I received) the latter is stored in float32 precision.

### Read times

The following are average times to read one slice from the NuLambda axis (i.e., rsp[i]).  Testing was on the same machine used for the conversion tests.

DC2 O3 continuum response: 4.5 ms (old), 6.4 ms (new)
DC3 O3 polarization response:  8.8 ms (old), 11 ms (new)
O4 version of continuum response: 350 ms (old), 46 ms (new)

"Old" is the previous response format, while "new" is the format in this PR.  Note that for large enough responses, such as the O4 continuum response, it's faster to read less data from disk and decompress than to read the data uncompressed.

By way of comparison, compressing the DC2 O3 continuum response with HDF5's gzip internal compression yielded a substantially *larger* file (889.77 MB) and read times of 21 ms, > 3x longer than with the BitShuffle compression used in this patch.


israelmcmc:
Thanks again @jdbuhler!

I'll check this PR in detail in a couple of weeks. In the meanwhile, I'd like ask the following people to take a quick look at specific parts of the code: 
- @GallegoSav: can you please take a look at the RspConverter? I know some features have been dropped by Jeremy, I'm OK with that. 
- @hirokiyoneda: can you please check the code impacting the imaging deconvoltution module and the extended source setup in FullDetectorResponse?
- @eneights: can you please check the parts of the code that use the "Pol" axis? Also, can you double check that your ASAD results are still consistent. Jeremy changed that expected values for the unit test, which were using a very very coarse response, but I would expect that the results from a not-so-coarse response are still consistent.

None of these individual checks need to be in-dept, just a quick look for things the pop to your eyes would be helpful.

cosipy/response/RspConverter.py
```diff
@@ -0,0 +1,901 @@
...
+        'NuLambda': "Location of the simulated source in the spacecraft coordinates",
+        'Pol': "Polarization angle",
+        'Em': "Measured energy",
+        'PsiChi': "Location in the Compton Data Space",
+        'Phi': "Compton angle",
+        'SigmaTau': "Electron recoil angle",
+        'Dist': "Distance from first interaction"
+    }
+
+    # ordered subset of .rsp axes to keep for HDF5 response
```
GallegoSav, line 70: is that important this order for hdf5 file ? Why not use the same as rsp ? It should not matter no ?

cosipy/response/RspConverter.py
```diff
@@ -0,0 +1,901 @@
...
+        ----------
+        rsp_filename: string
+           name of input file (must end with .rsp.gz)
+        h5_filename : string (optional)
+           name of output file (should end with .h5); if not
+           specified, use base name of rsp_filename with .h5 extension
+        overwrite : bool
+           overwrite the target filename if it exists
+        compress: bool
+           true iff response HDF5 file should use internal
```
GallegoSav, line 137: if

jdbuhler: That's "iff" as in "if and only if".

cosipy/response/RspConverter.py
jdbuhler: NuLambda being first is important pretty much everywhere, as rsp[i] is assumed to return a NuLambda slice.  In theory, we could change FullDetectorResponse.get_pixel() to accommodate arbitrary axis ordering, but for performance reasons, we might as well choose the order we are going to need later.  This is also the behavior of the previous conversion code.

cosipy/response/RspConverter.py
```diff
@@ -0,0 +1,901 @@
...
+
+                    edges = np.array_str(axis.edges.value,
+                                         max_line_width=100000000).strip("[] ")
+                    f.write(f"AD {edges}\n")
+
+            # write counts
+
+            f.write(f"StartStream {counts.size}\n")
+
+            # counts are printed one one line in FORTRAN order
```
GallegoSav: in one line

cosipy/response/RspConverter.py
jdbuhler: I will add that the old code put PsiChi before Phi in the HDF5 axis order, but nothing else in cosipy assumes one or the other ordering of these two axes. Where a particular ordering *is* needed, Phi comes first, and the existing cosipy code does a reordering of each CDS in memory if needed to ensure this. So I figured the on-disk format might as well match what everyone actually uses.

GallegoSav (commented):
Hi , I reviewed the code in the response folder. I just have minor comments about typo

cosipy/response/RspConverter.py
jdbuhler: Will fix...

cosipy/response/RspConverter.py
jdbuhler: One more note: I pushed a small change to verify that NuLambda is the first dimension of the response on disk when FullDetectorResponse opens the HDF5 file.  Removing this assumption would be a large undertaking, as many places in FullDetectorResponse assume that ._axes[1:] includes every dimension *except* NuLambda.  I don't believe there are any other such order dependencies in FullDetectorResponse besides this one.

israelmcmc:
Thanks @GallegoSav!

eneights:
The code involving the polarization axis looks good to me!

For the fit performed in the ICRS frame, the ASAD method results with this new response format are consistent with the previous results. For the fit performed in the spacecraft frame, the fitted polarization fraction is slightly inconsistent but improved (80% +/- 6% vs 91% +/- 7%, where the injected value was 80%), with a slightly lower MDP than before.

jdbuhler:
Do we think the output change is because I replaced the test response?  I would not expect the code itself to cause a different result.

eneights:
By the test response, do you mean the coarse response used for the unit test?

My comment was about the ASAD method notebook, which uses the full resolution polarization response.

jdbuhler:
Ok. Not sure why that result should have changed... I will check on it.

jdbuhler:
I just tried the tutorial here, and I get identical results for the polarization spacecraft fit with the develop and new_response branches (and the corresponding response files, which I just re-downloaded from wasabi):

Best fit polarization fraction: 0.803 +/- 0.061
Minimum detectable polarization: 0.16
Best fit polarization angle (IAU convention): 87.419 +/- 2.058

jdbuhler:
Is there any chance this could be a dependency issue?  I am using

 - python 3.10.16
 - numpy 2.2.6
 - numba 0.61.2
 - scipy 1.13.0
 - astropy 6.1.7
 - healpy 1.18.1
 - h5py 1.18.1
 - hdf5plugin 5.1.0
 - histpy 2.0.3.dev0 (the develop branch)

jdbuhler:
I wonder if the saved output in the polarization tutorial's .ipynb file is out of date even for develop?  When I ran the develop branch's version of it, I did not get either the assorted graphs or the run of error messages, and even the ICRS result was not the same as what was stored.

Stored ICRS output:

Best fit polarization fraction: 0.953 +/- 0.073
Minimum detectable polarization: 0.164
Best fit polarization angle (IAU convention): 87.668 +/- 1.987

My ICRS output (same for develop and new_response branches):

Best fit polarization fraction: 0.936 +/- 0.072
Minimum detectable polarization: 0.161
Best fit polarization angle (IAU convention): 87.668 +/- 1.987

So perhaps the difference is due to other, prior changes to develop, and the tutorial output was just never updated?

eneights:
@jdbuhler yes, it's definitely possible that the notebook output was just not updated, and that's what I was comparing with (I didn't rerun with the old response version). If you're getting identical results, I think it's good to go from my side

israelmcmc:
I went through this in more detail and all looks good to me. I'll go ahead an merge it. Thanks @jdbuhler 

Some closing notes:
- After seeing the new `to_dr()` function and how it uses DectectorResponse, I think we can update the DectectorResponse docstring to specify that it also works with an optional NuLambda axis, not just specific pixel. But we can do that in a separate PR.
- @jdbuhler: I'm planning on closing #199, #110 and #106, and add a checkmark for the response module in the list in #226. If you have a chance, can you please take a look at those issues and confirm?
- Thanks @eneights and @GallegoSav for taking a look!
- @hirokiyoneda Feel free to still take a look, but I think we can address any comments on a follow up PR. After a more detailed look, I didn't see any major changes to the imaging module, and I know you are pretty busy with other stuff.

jdbuhler:
@israelmcmc I agree on your proposed actions for open issues.

israelmcmc:
Thanks @jdbuhler

hiyoneda:
@israelmcmc @jdbuhler Sorry for reviewing this PR so late. While it was closed, I checked that this change does not affect the image deconvolution modules. I ran through the notebooks for 511 keV image deconvolution for both galactic and scatt methods, and confirmed that the results are consistent with previous ones.

israelmcmc:
No worries, @hiyoneda! And thank you for confirming, that's good to hear.

PR #367, opened 2025-06-02
Maximum likelihood polarization fitting

### LINKED ISSUE #211: Polarization fit using full likelihood in 3ML plugin.

eneights:
This includes updates to COSILike to allow for polarization to be fit using 3ML, as well as an example notebook which performs the fit in both the spacecraft and inertial frames. The fit currently either converges to incorrect values, or not at all. I still need to test this more to determine where the issue is coming from.

Closes #211


eneights:
Replaced by PR #443

PR #373, opened 2025-07-16
Speed improvements to point-source response computation from FDR

jdbuhler:
This patch modifies the FullDetectorResponse code to more efficiently compute point-source responses, especially for the ScattMap case. The main changes are:

 * vectorize rotation of PSRs from spacecraft-local to internal coordinates

 * avoid allocating Histograms until their contents are fully computed

 * minimizing use of high-overhead AstroPy coordinate arithmetic, and 
   vectorizing any remaining uses

There are a few other minor cleanups to the response directory and one tweak to polarization angle to support slightly more vectorization, which is used by the new response code.

# BENCHMARKS

(1) Using the DC3 O3 continuum response and the first hour of orientation data from the 1s DC2/DC3 orientation file, call get_extended_source() with nside_image=8 (earth occlusion is enabled by default).

OLD CODE: 6358 s
NEW CODE: 553 s (11.5x faster)

(2) Analogous to (1) using the DC3 polarization response (but we don't store the resulting PSRs because ExtendedSourceResponse does not support that).

OLD CODE: 65076 s
NEW CODE: 2035 s 2035 (32x faster)

# PERFORMANCE NOTES

In my profiling of benchmark #1, the new code spends about 43% of its time reading/computing slices of the response from disk, about 51% doing the math of rotating and adding PSRs, and the remaining 6% doing everything else (Attitude construction, coordinate transformations, find_bin).

The coordinate system for the Attitude, source, and per-PsiChi-pixel directions prior to rotation must all be the same (e.g., ICRS or galactic).  Currently, ICRS is used to match the output of the old code, though it would be preferable just to use source.frame (which is usually galactic). Switching coordinate systems should not change the output, but it does right now because the SpacecraftAttitudeMap's discretization of axes does not always yield orthogonal axis vectors. Once this issue is fixed in a future patch, and any expected test-case outputs are updated, we can start using source.frame.

If this code starts using interpolation rather than find_bin in the future, perhaps we should look into doing rotations in the spherical harmonic domain instead.

### DIFF
--- a/cosipy/polarization/polarization_angle.py
+++ b/cosipy/polarization/polarization_angle.py
@@ -94,8 +94,7 @@ class PolarizationAngle:
         pa_2 = Angle(np.arctan2(b, a), unit=u.rad)
 
         # Normalize the angle to be between 0 and pi
-        if pa_2 < 0:
-            pa_2 += Angle(np.pi, unit=u.rad)
+        pa_2 = np.where(pa_2 < 0, pa_2 + Angle(np.pi, unit=u.rad), pa_2)
 
         return PolarizationAngle(pa_2,
                                  self.skycoord,
--- a/cosipy/response/ExtendedSourceResponse.py
+++ b/cosipy/response/ExtendedSourceResponse.py
@@ -33,8 +33,9 @@ class ExtendedSourceResponse(Histogram):
         kwargs['track_overflow'] = False
 
         super().__init__(*args, **kwargs)
-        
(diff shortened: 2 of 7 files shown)


jdbuhler:
Missing lines in patch coverage are all error messages/exceptions

israelmcmc:
Thanks @jdbuhler! Nice work. The code looks to me and I verified that the the tutorials run fine (after fixing a couple of unrelated issues #374 #375  #376). I'm merging.

PR #378, opened 2025-09-05
SpacecraftFile improvements, including new scatt_map and interpolation methods

jdbuhler:
This patch addresses two computation issues in SpacecraftFile .

(1) The SpacecraftAttitudeMap is currently stored as a Histogram that discretizes the x and y (or z) axes of the Attitudes represented by the SpacecraftFile.  Because the axes are discretized separately, they do not maintain orthogonality.  Turning them back into an Attitude, as is done in FullDetectorResponse.get_point_source_response(), therefore has to apply some arbitrary distortion (done inside the Rotation class) to make them orthogonal again.

We address this issue by (a) making SpacecraftAttitudeMap hold arrays of Attitudes + weights, rather than using the intermediate Histogram, and (b) discretizing the rotvec representation of Attitude instead of its axes representation, which can be done without compromising orthogonality.  The impact of this change on observed outputs from cosipy is, to the best of my knowledge, quite small.  We introduce a SpacecraftAxisMap class that preserves the old functionality in the rare cases where code was dependent on the Histogram representation -- primarily for plotting spacecraft paths.

(2) The code for extracting a time range from a SpacecraftFile, namely source_interval(), uses linear interpolation of axes at the start and end times, which again does not preserve orthogonality.  To address the underlying representation concern, (a) the set of Attitudes for the SpacecraftFile is now created at class initialization time and is used as the primary representation in all subsequent code; (b) the interpolation is now done directly on the Attitudes using SLERP on their quaternion representations.  We also use vector SLERP to interpolate earth pointings, since it is a "constant angular velocity" interpolation method that better achieves the goal of interpolation here.

In addition to the above changes, the patch makes a number of minor API cleanups and performance improvements to SpacecraftFile.  All of these should be compatible with Israel's ongoing PR 369 changes (though that patch goes further in reorganizing the API). 

It should be noted that the frame passed when constructing SpacecraftFile is now used consistently for all stored directions as well as the Attitudes, including the Attitudes passed off to the scatt_map.  This means that we no longer have to go through ICRS in the scatt_map case of FDR.get_point_source_response() -- everything is in the galactic frame to begin with.

The one invasive API change that I did not attempt in this patch is to make get_dwell_map() take a HealpixBase rather than always reading it from a response file.  This is easy to do (and is done in PR 369) but requires work at all the call sites to hoist the file opening out of loops, so we don't have to do it every time we create a dwell map.  I also did not attempt to separate out the XSPEC code into its own class as PR 369 does, but I did ensure that it no longer depends on implicit caching of state by the main SpacecraftFile functions.


cosipy/spacecraftfile/SpacecraftFile.py
```diff
@@ -263,380 +297,436 @@ def source_interval(self, start, stop):
...
+        angle_axis = Axis(np.linspace(0., 2*np.pi, num=nside*8+1), unit=u.rad)
+
+        r_lon, r_colat = self._cart_to_polar(rot_dirs)
+        dir_bins = dir_axis.find_bin(theta=r_colat.value,
+                                     phi=r_lon.value)
+        angle_bins = angle_axis.find_bin(rot_angles)
+
+        dv = dir_axis.pix2vec(dir_bins)
+        av = angle_axis.centers[angle_bins]
+        att1 = Attitude.from_rotvec(np.column_stack(dv) * av[:,None],
```
israelmcmc: I think these 3 lines are leftover from testing. They don't seem to be used.

cosipy/spacecraftfile/SpacecraftFile.py
```diff
@@ -263,380 +297,436 @@ def source_interval(self, start, stop):
...
-            weight[earth_occ_index[:-1]] = 0
-        
-        # Fill histogram:
-        h_ori.fill(x, y, weight = weight)
+        # remove any attitudes with zero weight
+        binned_attitudes = binned_attitudes[binned_weights > 0]
+        binned_weights   = binned_weights[binned_weights > 0]
 
-        return h_ori
+        return SpacecraftAttitudeMap(binned_attitudes,
```
israelmcmc, line 707: I was thinking that  SpacecraftAttitudeMap will be a histogram, optionally a sparse one, with the weight being the contents. Did you store the attitudes directly so you don't have to recompute them based on the axes value later in the process? If so, that's fine.

israelmcmc: Also, eventually the SC will likely go through all possible (binned) attitudes. I think we can handle it even in a list, but just so you are aware. DC3 had a simplified orbit and attitude evolution

cosipy/spacecraftfile/SpacecraftFile.py
```diff
@@ -263,380 +297,436 @@ def source_interval(self, start, stop):
...
+        # Get orientations as rotation vectors (center dir, angle around center)
+
+        rot_vecs   = self._attitude[:-1].as_rotvec()
+        rot_angles = np.linalg.norm(rot_vecs, axis=-1)
+        rot_dirs   = rot_vecs / rot_angles[:,None]
+
+        # discretize rotvecs for input Attitudes
+
+        dir_axis = HealpixAxis(nside=nside, coordsys=self.frame)
+        angle_axis = Axis(np.linspace(0., 2*np.pi, num=nside*8+1), unit=u.rad)
```
israelmcmc: The angle binning can be an optional parameter. nside*8+1 seems like a good default though.

cosipy/spacecraftfile/SpacecraftFile.py
israelmcmc: I also wonder if it would be worth to change the class name, since it totally changed. It's not a map anymore, and the API is different. What do you think? SpacecraftAttitudeList?

cosipy/test_data/image_deconvolution/all_sky_image_model_test_nside1.hdf5
israelmcmc, line 1: I'm not sure why this or any of the files in image_deconvolution changed? Did all of these files change because of the attitudes being slightly different?

israelmcmc (commented):
Thanks @jdbuhler! And again sorry for the delay. This looks good to me, and great improvement overall. I left a few comments and questions.

israelmcmc:
Oh, and there seems to be some conflicts now. I haven't checked that, but that's why the unit tests are not running.

cosipy/spacecraftfile/SpacecraftFile.py
jdbuhler: Added optional parameter to scatt_map constructor to allow the number of angle bins to be changed

cosipy/spacecraftfile/SpacecraftFile.py
jdbuhler: Yes, this was debug code.  Removed.

cosipy/test_data/image_deconvolution/all_sky_image_model_test_nside1.hdf5
jdbuhler: I regenerated the test data after implementing the changes to orientation interpolation and scatt_map construction.  I assumed that at least some of it might depend on these computations.  But I can't tell what represents actual changes in the values and what might just be timestamp changes or similar in the HDF5. The timestamps get get updated every time an HDF5 file is recreated, even if the content is exactly the same.

cosipy/spacecraftfile/SpacecraftFile.py
jdbuhler: As described above, SpacecraftAttitudeMap is now an array of (discretized) Attitudes and corresponding array of weights.  This is more or less what was represented by the previous sparse Histogram, except that we store the actual Attitudes rather than trying to represent them (with extra rounding error) as a pair of HEALPixels.

For sufficiently long time periods, the spacecraft will go through every Attitude. But for short periods, it is likely to have just a few different Attitudes.  I don't think there is any performance benefit to dropping the sparse representation even if every Attitude is eventually visited.

I was hesitant to change the class name because that change would touch a lot of other source files, but if you'd prefer, we can call it something new.

jdbuhler:
Conflicts fixed -- merged with develop as of about 9 AM on 10/22/25

cosipy/spacecraftfile/SpacecraftFile.py
jdbuhler: I would probably just name the class SpacecraftAttitudes (scatts for short).  There's no need to specify how the attitudes are stored as part of the class name.

jdbuhler: You could perhaps more appropriately call it SpacecraftBinnedAttitudes, since the original SpacecraftFile stores an attitude per time point, while this class stores a time-weighted set of attitudes but has no notion of the time at which each attitude was assumed.

jdbuhler: While there are just a few places that use the SpacecraftAttitudeMap name, there are *many* places that use the term "scatt_map" as a part of a parameter or function name.  If we change the class name, I would likely replace all occurrences of "scatt_map" with "scatts".

cosipy/test_data/image_deconvolution/all_sky_image_model_test_nside1.hdf5
israelmcmc: I checked if the test would still work if these were reverted. They fail. I suppose that the small changes in attitude result in this change. Hopefully, we can fix this issue in the test suite in the future.

PR #385, opened 2025-09-09
Deprecate TSMap in favor of FastTSMap

israelmcmc:
FastTSMap is the one being used and maintained now. I think we should change it's name back to simply "TSMap", but for now let's remove the old code to clean things up.

I confirmed that it was no longer used by any other piece of code or tutorial in cosipy.

Well, except for a few imports in the imaging tutorials that I don't think were necessary, so I removed them. @hirokiyoneda please confirm.


Yong2Sheng:
Hi @israelmcmc @hirokiyoneda , I checked that `TSMap` is only imported but not used in the image tutorials. So i think we are safe to merge.

Yong2Sheng (approved):
No issue found, merging now!

israelmcmc:
Thanks @hirokiyoneda and @Yong2Sheng!

PR #404, opened 2025-09-15
Pass `cpu_cores` para correctly for MOCTSMap.moc_ts_fit function

Yong2Sheng:
A tiny fix to function `MOCTSMap.moc_ts_fit function` so that the number of CPU cores requested is passed to the parallel fit. Thanks @NicoloParmiggiani for finding this bug!


avalluvan (approved):
The change is very minor. However, no tests are covered through codecov so that did not run successfully.

PR #416, opened 2025-10-06
Earth occultation in ScAtt + Local image analysis

### LINKED ISSUE #381: Account for Earth's occultation in RL (scatt bins approach)
See #96 

COSILike already accounts for Earth occultation, as well as the RL deconvolution when using Galactic coordinates.

The RL deconvolution when using local coordinate / scatt bin does not yet account for this.

hiyoneda:
Following PR #381 , the Earth occultation is now considered in image analysis using local coordinate / scatt bin.

Basically, there are only two major changes:
1. Now SpacecraftAttitudeExposureTable keeps the information about the Earth zenith and altitude of the satellite, which will be used for the Earth occultation calculation.
2. In CoordsysConversionMatrix. spacecraft_attitude_binning_ccm, exposure time maps for all scatt bins are calculated. In this step, the Earth occultation is considered, and exposure time is set to zero if the Earth blocks a corresponding pixel.

Additionally, there are minor changes.
- These methods are removed because they are not used anymore.
    - CoordsysConversionMatrix. time_binning_ccm
- These methods are added
    - CoordsysConversionMatrix.calc_exposure_map: It can calculate an exposure map by convolving the exposure time map with the effective area for each scatt bin
    - SpacecraftAttitudeExposureTable.get_binned_data_scatt: It can generate a binned histogram from an event file. The generated histogram has 'ScAtt' in its axes. Initially, I planned to add it to the DataIO class, but considering that this approach (ScAtt + local) is not a default analysis, I decided to put it in SpacecraftAttitudeExposureTable to avoid confusion in the DataIO class.


GallegoSav (approved):
I tested with DC3 response and files and it worked well

PR #418, opened 2025-10-10
Adding the input file name as an option in the tra file reader function

GallegoSav:
I added an option to pass directly the input file name to the tra file reader instead of using only the one define in the yaml file.

This is more convenient when you have a list of files to read and you don't want to change the yaml file each time.


ckarwin (approved):
I tested the new input option. The code works for me with and without passing an input file. The logic also makes sense, and the changes are very straight forward. This can be merged.

PR #369, opened 2025-06-30
Interfaces refactoring

israelmcmc:
Work in progress...

This is a major refactoring of cosipy to make it more modular, using interfaces (protocols) to decouple the inputs/outputs of different classes from their implementation.

There are currently three examples that serve as an entry point to see the interfaces usage:
- [docs/api/interfaces/examples/toy](https://github.com/israelmcmc/cosipy/tree/interfaces/docs/api/interfaces/examples/toy): A generic 1D fits of a Gaussian signal on top of a flat background
- [docs/api/interfaces/examples/grb](https://github.com/israelmcmc/cosipy/tree/interfaces/docs/api/interfaces/examples/grb): Equivalent to the [GRB spectral fit tutorial](https://github.com/cositools/cosipy/blob/develop/docs/tutorials/spectral_fits/continuum_fit/grb/SpectralFit_GRB.ipynb)
- [docs/api/interfaces/examples/crab](https://github.com/israelmcmc/cosipy/tree/interfaces/docs/api/interfaces/examples/crab): Equivalent to the [Crab spectral fit tutorial](https://github.com/cositools/cosipy/blob/develop/docs/tutorials/spectral_fits/continuum_fit/crab/SpectralFit_Crab.ipynb)

Presentations:
- 
[20250630-CosipyInterfacesUpdate-Israel.pdf](https://github.com/user-attachments/files/21005795/20250630-CosipyInterfacesUpdate-Israel.pdf) (at version 214f2ae9b63a574074f)

Installation notes and troubleshooting:
-
- You need to install this histpy versions from source (branch time_axis): https://gitlab.com/burstcube/histpy/-/tree/time_axis
- need to copy `crab.yaml` and `background.yaml` in the repository or add it on wasabi in `cosipy_tutorial`

TODO list:
-
- [x] Binned analysis toy example
- [x] GRB spectral fit (using dwel time map / local coordinates)
- [x] Crab spectral fit (using scatt map / galactic coordinate)
- [x] Unbinned analysis toy example
- [x] Extended source response
- [x] Propagate to source injector
- [x] Propagate to TS map
- [x] Propagate to imaging deconvolution
- [x] Merge with Jeremy's and other recent PRs
- [x] Update all notebooks
- [x] Fix all tests

PR #424, opened 2025-10-23
Rewrite of polarization ASAD analysis

jdbuhler:
The primary goal of this PR is to rewrite polarization ASAD analysis for better computational performance. It includes the following optimizations:

1. The existing ASAD calculations from unbinned data, binned data, or a response plus a scatt_map or dwell map have been vectorized. Methods of PolarizationAngle used by these operations have also been vectorized.

2. The computation of ASADs has been reorganized to remove redundant operations. In particular, computing the polarized and unpolarized ASADs from the response no longer repeatedly computes the same set of scattering directions.

3. The source spectrum, which does not change during the computation, is integrated once before constructing the ASADs, rather than repeatedly each time the integrated flux is needed.  This change required extending PointSourceResponse.get_expectation() to take an optional pre-integrated flux array, rather than always doing the spectrum integration itself.

A further optimization was performed on the Band_Eflux spectral model used with polarization analysis in the units tests and the tutorial notebook.  Evaluating this model at a given energy x and set of spectral parameters numerically integrates a Band spectrum between specified limits [a,b] to compute a normalizing constant. Evaluation for each new x recomputes this integral, even if the spectral parameters have not changed since the previous evaluation. We now cache the spectral parameters and integral in the model so that successive evaluations with the same parameters can simply reuse the cached integral.  This change is transparent to the user and should not impact the model's use in, e.g., spectral fitting.

In the course of vectorizing the PolarizationAngle code, it was necessary to ensure that the source direction was always a scalar rather than a vector SkyCoord.  This prompted a change to the point source response calculation in FullDetectorResponse, which assumes that the source direction is a vector SkyCoord. Passing a vector with multiple source directions was done as an optimization in the past, but it is no longer effective after a recent rewrite of FullDetectorResponse, nor is it even possible in the physically relevant case that earth occultation is considered in the scatt_map. This patch therefore requires that get_point_source_response() takes a scalar source, which created an opportunity to simplify its code. The DetectorResponse.ipynb notebook was updated to remove references to vectorized sources.

There are also some other minor optimizations in the response directory.

One optimization that was **not** done was modifying the dwell-time map construction to use a preloaded FullDetectorResponse rather than reopening the response file each time. This change is part of Israel's pending PR #369.

### PERFORMANCE IMPACT

For the ASAD analysis on unbinned inputs in the galactic frame that is done in the tutorial notebook, the new code runs roughly 136x faster than the version on the development branch.

 - old code: 158.96s
 - new code: 1.164s

More than half the time spent in the new code is in the single required integration of the Band_Eflux spectrum.

There is just one change that meaningfully impacts the result of the analysis. The scaled background ASAD was previously truncated to integer values rather than being left as floating-point after scaling. Removing this truncation (the reason for which is unclear) changes the predicted PA, PF, and MDP in the tutorial notebook by at most one part in a thousand versus the old code. Without this change, the old and new code agree to within floating-point precision (around one part in 10^7).

### TEST SUITE

The test suite has been extended to exercise the vectorized PolarizationAngle operations and to check the output of the ASAD computation, including for the case of binned data (which was previously not covered and was in fact broken).


cosipy/response/PointSourceResponse.py
```diff
@@ -54,7 +54,9 @@ def get_expectation(self, spectrum, polarization=None):
             Spectral hypothesis.
         polarization : 'astromodels.core.polarization.LinearPolarization', optional
             Polarization angle and degree. The angle is assumed to have same convention as point source response.
-        
+        flux : np.ndarray of float, optional
```
israelmcmc: I think this is a Quantity or a Histogram with units, isn't it? Flux should have units of [1/area/energy/time] (3ML defaults to 1/cm2/s/keV) Later in the function flux.unit is called.

cosipy/polarization/polarization_angle.py
```diff
@@ -46,69 +51,76 @@ def convention(self):
...
     def from_scattering_direction(cls, psichi, source_coord, convention):
         """
         Calculate the azimuthal scattering angle of a scattered photon.
-        
+
         Parameters
         ----------
         psichi : astropy.coordinates.SkyCoord
-            Scattered photon direction
+            Scattered photon direction FIXME: could be a vector skycoord
```
israelmcmc: I think it does work for a vector skycoord, but I haven't stressed it. Did you find a case where it fails?

israelmcmc (commented):
Thanks @jdbuhler. This is what I have so far. I still need to run it.

cosipy/response/PointSourceResponse.py
jdbuhler: You're right -- it is a Histogram.  Fixed the docstring

cosipy/polarization/polarization_angle.py
jdbuhler: The FIXME comment was stale -- removed.  psichi can indeed be a vector

PR #427, opened 2025-10-23
Make the adding of CO_seq in the data files retrocompatible

GallegoSav:
Now we check the number of keys in the unbinned file before writing it


jdbuhler:
Reviewed.  It works correctly for both the self-test (new-style input file) and the polarization notebook (old-style input file).
Approved.

Once all the tutorial inputs have been updated to the new form (prior to DC4?), support for the old format should be retired.

jdbuhler (approved):
Reviewed and verified that it works with both the unit tests and the polarization notebook

PR #429, opened 2025-10-24
Fix automated running of ts_map notebook

jdbuhler:
Recent changes to develop renamed the tutorial notebook associated with ts_map.  This patch updates the run_tutorials.yml database to point to the correct notebook file so that run_tutorials.py can run it.


israelmcmc:
thanks, @jdbuhler!

PR #439, opened 2025-11-06
Minor typos

augustus-thomas:
While reading/linting cosipy I found 3 minor typos:

* COSILike using "log.error" instead of "logger.error"
* RspConverter raising "RunTimeError" instead of "RuntimeError"
* RspConverter._get_eff_area using "e_low" instead of "e_lo" in the powerlaw case

This PR fixes them.


GallegoSav (approved):
nice catch of the typos  ! I merge this

PR #440, opened 2025-11-07
Adding time binning for the image deconvolution using the local coordinate, with code refactoring

hiyoneda:
#### 1. ExposureTable base class refactoring

- Created a new base class `ExposureTableBase` to provide a flexible framework for different exposure table calculation methods
- Refactored existing `SpacecraftAttitudeExposureTable` as a child class of `ExposureTableBase`
- This design allows for easy implementation of different exposure calculation strategies (e.g., time-binned)

#### 2. Critical bug fix: Proper SAA handling

Previous implementation used simple time width instead of livetime, which **did not properly account for South Atlantic Anomaly (SAA) passages**.

- Modified exposure calculations to use **livetime** 
- SAA cut periods are now correctly excluded from exposure calculations

#### 3. Time-binned image deconvolution with local coordinates

This feature restores and improves functionality that previously existed in the coordinate conversion matrix.

- **`TimeBinnedExposureTable`**: New exposure table class that handles multiple time bins
  - Calculates spacecraft attitude information for each time bin

- **`CoordsysConversionMatrix`**:
  - The `from_exposure_table()` method now accepts all child classes of `ExposureTableBase`

#### 4. Tutorial organization

- Moved the time-binned local coordinate tutorial to `image_deconvolution/optional/`
- **Rationale**: This is NOT the default analysis approach for COSI
- Prevents user confusion by clearly separating optional methods from standard workflows


hiyoneda:
download-1
download

The comparison between the reconstructed 511 keV maps before/after fixing the bug about the SAA handling. The top is a map with the previous code, and the bottom is the new one. The injected model is the NFW model.

The reconstructed fluxes are 8.7e-4 cm-2 s-1 for the previous one and 1.37e-3 cm-2 s-1 for the updated code.

The injected model's flux is 1e-3 cm-2 s-1, which is 40% higher than the result with the new code.

hiyoneda:
The response file I used for ScAA+Local appears to be incorrect. By modifying as using `Response511.o4.e509_513.s20881894470591.m2555.filtered.nonsparse.binnedimaging.imagingresponse_nside16.area.good_chunks.h5`, the results seem consistent with the analysis with galactic CDS. Especially, the exposure maps are consistent within a few 0.1% differences, and the SAA is correctly removed. 

Once the response file used in the notebook is modified, this PR is ready to be reviewed.

Screenshot 2025-11-10 at 19 58 29

hiyoneda:
One thing to be resolved is why the flux from the NFW model is twice that of the injected model (according to the source file of this simulation, the total flux should be 1e-3). In both cases (galactic or scat+local), there is the same discrepancy, so it looks like any issue in the simulated event or the original response.

hiyoneda:
> One thing to be resolved is why the flux from the NFW model is twice that of the injected model (according to the source file of this simulation, the total flux should be 1e-3). In both cases (galactic or scat+local), there is the same discrepancy, so it looks like any issue in the simulated event or the original response.

@ckarwin, do you have some thoughts about it by chance? It may be better for me to investigate this issue with the DC4 dataset.

GallegoSav:
@hiyoneda Could you make the Earth occultation optional (by default set to True)[here](https://github.com/hiyoneda/cosipy/blob/2da9d59cf62e589b4265a9ccd30bf1cd1171d152/cosipy/image_deconvolution/coordsys_conversion_matrix.py#L148) so that I can use it for the balloon analysis.

Something like

```
for i_pointing in range(num_pointings):
            if Earth_occ :
                earth_radius = np.pi - np.arcsin(r_earth / (r_earth + altitude[i_pointing])) #rad
                filling_pixel_index = hp.query_disc(nside_model, hp.ang2vec(earth_zenith[i_pointing,0], earth_zenith[i_pointing,1], lonlat = True), nest = is_nest_model, radius = earth_radius)
                exposure_time_map[i_pointing][filling_pixel_index] = livetime[i_pointing]
            else :
                exposure_time_map[i_pointing, :] = livetime[i_pointing]

```

hiyoneda:
> @hiyoneda Could you make the Earth occultation optional (by default set to True)[here](https://github.com/hiyoneda/cosipy/blob/2da9d59cf62e589b4265a9ccd30bf1cd1171d152/cosipy/image_deconvolution/coordsys_conversion_matrix.py#L148) so that I can use it for the balloon analysis.
> 
> Something like
> 
> ```
> for i_pointing in range(num_pointings):
>             if Earth_occ :
>                 earth_radius = np.pi - np.arcsin(r_earth / (r_earth + altitude[i_pointing])) #rad
>                 filling_pixel_index = hp.query_disc(nside_model, hp.ang2vec(earth_zenith[i_pointing,0], earth_zenith[i_pointing,1], lonlat = True), nest = is_nest_model, radius = earth_radius)
>                 exposure_time_map[i_pointing][filling_pixel_index] = livetime[i_pointing]
>             else :
>                 exposure_time_map[i_pointing, :] = livetime[i_pointing]
> ```

Yes, the option has been added.

cosipy/image_deconvolution/exposure_table_base.py
```diff
@@ -0,0 +1,383 @@
...
+            List of pointings in degrees, e.g., [ np.array([l, b]), np.array([l, b]), ...]
+        livetime : list of float
+            List of livetime in seconds for each pointing, e.g, [ 1.0, 1.0, ...] 
+
+        Returns
+        -------
+        :py:class:`np.array`
+            Averaged pointing in degrees, as np.array([l, b])
+        """
+        if np.sum(livetime) == 0:
```
GallegoSav: Hi @hiyoneda for the case of the Balloon analysis , sometime for a specific time bin, the livetime is = 0 .
But then during the CoordsysConversionMatrix computation, it is raising errors because it tries to manipulate None object. 
Could you add a check during the conversion matrix part where you don't take into account None Average pointing ? 

For me I changed that to return 0 instead. it will give a wrong pointing but it does not really matter as the livetime is 0 so it won't play a role in the exposure

cosipy/image_deconvolution/time_binned_exposure_table.py
```diff
@@ -0,0 +1,234 @@
...
+        Returns
+        -------
+        :py:class:`histpy.Histogram`
+            Binned data with axes ["Time", "Em", "Phi", "PsiChi"].
+        """
+        # Get TimeTags from unbinned event
+        time_tags = unbinned_event.cosi_dataset['TimeTags']
+
+        # Warning about assumptions
+        logger.warning("This method assumes that TimeTags are sorted in ascending order.")
```
GallegoSav, line 180: Actually for simulations, the TimeTags are often shuffle since we merge different sim files together. Instead of just printing a warning maybe there could be a check if the timetag is in ascending order or not.

If not we can sort the dict before doing anything and raise the log warning

GallegoSav (approved):

cosipy/image_deconvolution/time_binned_exposure_table.py
hiyoneda: Memo: It will be resolved by #455

cosipy/image_deconvolution/exposure_table_base.py
hiyoneda: It is modified as to return an averaged pointing with equal weight when livetime is all zero.

GallegoSav (approved):

israelmcmc:
Thanks @GallegoSav and @hirokiyoneda !

PR #442, opened 2025-11-12
Add the user option of weight in get_binned_data

GallegoSav:
I added the `weights` argument in `get_binned_data` so that now you can specify it when filling the `Histogram`.

If not set, histpy will choose 1 , like it is done for now. 

I tried and it worked for me because I need it for the Balloon analysis

### DIFF
--- a/cosipy/data_io/BinnedData.py
+++ b/cosipy/data_io/BinnedData.py
@@ -20,7 +20,7 @@ class BinnedData(UnBinnedData):
 
     def get_binned_data(self, unbinned_data=None, output_name=None, \
             make_binning_plots=False, show_plots=False, \
-            psichi_binning="galactic", event_range=None):
+            psichi_binning="galactic", event_range=None,weights=None):
 
         """Bin the data using histpy and mhealpy.
         
@@ -41,7 +41,10 @@ class BinnedData(UnBinnedData):
             'local' for binning in local coordinates. Default is Galactic. 
         event_range : list of integers, optional 
             min and max event to use for the binning. 
-        
+        weights : value or array of values, optional
+            weight to use when filling the Histogram, if not set histpy will 
+            use weight of 1
+            
         Returns
         -------
         binned_data : histpy:Histogram
@@ -114,14 +117,14 @@ class BinnedData(UnBinnedData):
             self.binned_data.fill(self.cosi_dataset['TimeTags']*u.s, 
                     self.cosi_dataset['Energies']*u.keV, 
                     np.rad2deg(self.cosi_dataset['Phi'])*u.deg, 
-                    coords)
+                    coords,weight=weights)
         if event_range != None:
             low = int(event_range[0])
             high = int(event_range[1])
             self.binned_data.fill(self.cosi_dataset['TimeTags'][low:high]*u.s, 
                     self.cosi_dataset['Energies'][low:high]*u.keV, 
                     np.rad2deg(self.cosi_dataset['Phi'][low:high])*u.deg, 
-                    coords[low:high])
+                    coords[low:high],weight=weights)
 
         # Save binned data to hdf5 file:
         if output_name != None:


ckarwin (commented):
Hi @GallegoSav,
I looked at the code and the changes look fine. As a sanity check, I also verified that the dataIO tutorial runs ok. Finally, I tested filling a histogram with the binned dataIO class and it works. 

Out of curiosity, why do you need a weighted histogram for the balloon data?

One request: the weight can be a single value or an array. For a single value, the weight is applied to all bins. This is straight forward to understand. For an array, it's a bit more complicated b/c the weight needs to be specified explicitly by key. Do you think it would be worth documenting the requirements to pass an array as a weight? This could either be a comment in the doc string (e.g. specifying the required array size) or maybe even a simple example in the dataIO notebook. But I'm not sure how common it will be for people to use a weighted histogram, so maybe an explicit example is not needed. Please let me know what you think.

GallegoSav:
> Hi @GallegoSav, I looked at the code and the changes look fine. As a sanity check, I also verified that the dataIO tutorial runs ok. Finally, I tested filling a histogram with the binned dataIO class and it works.
> 
> Out of curiosity, why do you need a weighted histogram for the balloon data?
> 
> One request: the weight can be a single value or an array. For a single value, the weight is applied to all bins. This is straight forward to understand. For an array, it's a bit more complicated b/c the weight needs to be specified explicitly by key. Do you think it would be worth documenting the requirements to pass an array as a weight? This could either be a comment in the doc string (e.g. specifying the required array size) or maybe even a simple example in the dataIO notebook. But I'm not sure how common it will be for people to use a weighted histogram, so maybe an explicit example is not needed. Please let me know what you think.

Hi @ckarwin , 

I asked this question in the past to Israel, I am putting his answer below : 

Nope, it’s simpler than that. fill()  doesn’t read a file, it just takes arrays of values, one for each axis. They can come from anywhere, but if they are from fits, you’ll have to read the file some other way and convert them to arrays. What this comment means is purely about the python syntaxis, though. If, for example, your histogram has two axes, you need to use the following:

`h.fill(x,y,weights = w)`

Where x, y and w are all arrays. The following won’t work:

`h.fill(x,y,w)`

The explicit use for `weights =` is what the comment is warning about.

This is exactly how you use `np.histogram` or `ax.hist` , you can give a value or an array but you need to specify `weights=`, which is done in `the get_binned_data`.  For my case I used a value.


I need it for the power law correction for the background but also when I have a data driven background I want to re-normalized in  the signal region.

ckarwin:
@GallegoSav 

> I need it for the power law correction for the background but also when I have a data driven background I want to re-normalized in the signal region.

Ok, I see. 

Regarding the weights, it's still not clear to me how I would pass an array of values. Can you please provide me a simple example that I can test (i.e. how do I define the weights)? I understand that this is essentially a histpy question that the user would need to figure out, but if we can provide some basic guidance it would be a good thing. I would also like to verify that it can be readily done in cosipy if the option will be provided.

GallegoSav:
@ckarwin  I would do like this :
```
from cosipy import BinnedData
import numpy as np

analysis = BinnedData("inputs-yaml") #set the path of the unbinned file in the yaml config file
data = analysis.get_dict() #get the data

#create an array of weights. Length must be = nb of evt !!!
weights = np.full(len(data),3) # here I choose to set the weight to 3 for every event 
                                                     #but it could be any values and different for each event

#fill the histogram
analysis.get_binned_data(unbinned_data = path_file, psichi_binning="galactic",weights=weights)

```

ckarwin:
Hi @GallegoSav 

That's not exactly what I was looking for. And actually, I don't think that works b/c `len(data)=12` since it's the cosi dataset which has 12 keys. Anyways, I don't think it's too important for this PR. It's a more a problem of how to use the weight keyword in histpy, and this can be figured out by the user following the histpy documentation and examples.

ckarwin (approved):

israelmcmc:
Thanks for working on this @GallegoSav and for reviewing @ckarwin .

Looking at the conversation, I think the confusion comes because the `weights` are applied to each event, not to the bins. It can be either a single value, or an array with size equal to the number of events. In hitspy it's more obvious because it's used in the `fill()` method, in order to create a weighted histogram.

Issue #414, opened 2025-10-01
COSILike Class Object in COSI 3ML Gives AttributeErrors when calculating likelihoods


orionning676:
After making a COSILike Class object (screenshot 1) following the diffuse 511 keV tutorial ( cosipy/docs/tutorials/spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb ) , attempting to retrieve a simple log likelihood results in an attribute error, which looks like an internal bug (screenshot 2).

<img width="903" height="368" alt="Image" src="https://github.com/user-attachments/assets/90eee149-5396-4b5d-af8b-d74c8ff9ecbf" />

<img width="2206" height="558" alt="Image" src="https://github.com/user-attachments/assets/b67553d6-393b-4e8a-919f-fccc34b6974a" />

israelmcmc:
Thanks for reporting, @orionning676! Which cosipy version are you using? Are you running it from a released version or from source?

orionning676:
Hi Israel, I am using cosipy version 0.4.0dev0, and I ran it from source.

Thanks!

israelmcmc:
Thanks! Since you are running it from source, can you please run `git rev-parse HEAD` to get the specific commit? I'm assuming you installed it with `pip install -e .`. If you didn't use `-e`, then I'd need the commit you had when you last run `pip install .` (or you can `pip install -e .` again and see if you still see the error)

orionning676:
This seems to be the commit: 

e9dca56e9b808a1e42cd9f21da36d4d826bb0d72

israelmcmc:
@orionning676 Sorry for the delay. I think #446 fixed this. Please update your develop branch and give it another try. If it's working now, please close the issue. Also, if you are not doing this already, don't forget to call set_model() before get_log_like().

israelmcmc:
Actually, I'll close this issue to help with the clutter, but please reopen it if you still see the error.

PR #446, opened 2025-11-25
Make sure _printed_warning is set by moving it to the __init__

israelmcmc:
For #414


israelmcmc:
Given our PR backlog and that this is a maintenance task, I'll bypass the review and directly merge it.

Issue #445, opened 2025-11-23
galactic-frame response files still have track_overflow enabled


jdbuhler:
The galactic-frame response in COSI-SMEX/cosipy_tutorials/ts_maps/Albedo_galactic_CDS_binned.hdf5 was created before histpy was modified not to add overflow tracking to each dimension of Histograms by default.  The extra padding is not needed but requires extra code where the response is used (e.g., in ts_map) to account for it.

For DC4, we should fix this file and any other precomputed responses to strip the overflow padding, unless there is a good reason to keep it.

jdbuhler:
Assuming my PR #447 is accepted, the partial support therein for padded galactic-frame responses should be removed once we resolve this issue.

PR #447, opened 2025-12-02
Performance engineering for ts_map

jdbuhler:
Cosipy's code to map the location of transients in the sky (FastTSMap and MocTSMap) is one of the more challenging computations in the library.  It compares an observed set of Compton events to a response-derived model for each of hundreds to thousands of _different_ source (NuLambda) directions. Each galactic-frame source direction is typically converted to a path in the spacecraft frame, which is in turn discretized into an exposure map that is used to retrieve and weight multiple slices of the instrument response, forming a model that is fit, along with the background model, to the data to obtain a likelihood score. These operations heavily exercise multiple parts of the library (SpacecraftFile, FullDetectorResponse, and PointSourceResponse in particular) in a way that highlights performance issues.

This PR implements a number of performance improvements that collectively speed up likelihood mapping of transients by one to two orders of magnitude.  Many of the ideas are drawn from the work in [my recent ICRC abstract](https://sbs.wustl.edu/pubs/ICRC2025_587.pdf), but unlike that work, this PR does _not_ require the entire response to be read into memory prior to mapping.

### Key Performance Improvements

(1) fast_norm_fit has been vectorized, accelerated, and parallelized using Numba JIT compilation. (Numba is already a dependency of histpy and therefore of cosipy, so this change does not introduce any new library dependencies.)

(2) PSR computation from the response has been reorganized to remove redundant computation and exploit the inherent _sparsity_ of the fitting computation for short transients. In particular, when mapping with only a limited number of Compton events, most voxels of the CDS defined by the response are empty.  Reducing the computation to just the nonempty voxels as soon as possible substantially reduces the amount of arithmetic needed. Similarly, projecting away the Em dimension and convolving Ei with the spectrum as soon as possible results in less computation later on.

fast_norm_fit has been modified to take sparsified data, background, and excess vectors, along with enough extra information (the sum of the excess vector prior to sparsification) to produce the same result as the prior code.  Iteration can therefore skip CDS voxels that would contribute zero to the ts and its derivatives.

(3) Access to PSRs has been wrapped in a caching mechanism. While the PSR computations needed to compute an nside=16 map retrieve and process over 15000 response slices from disk, the current nside=8 response has only 768 distinct slices. Even for a finer-grained response, nearby source pixels frequently translate to local-frame paths that reference some of the same response slices. Reading a slice from a FullDetectorResponse or a galactic-frame response on disk is inherently slow, as it incurs OS overhead, decompression overhead for FDRs, and overhead associated with the h5py library. Caching these slices removes most of the retrieval overhead when they are used multiple times.

While it is possible to configure h5py's chunk cache to keep some response slices in memory (and I've added a bit of code to do so if desired in FullDetectorResponse), this approach doesn't address the overhead incurred simply by calling into h5py, or the overhead of processing each retrieved response slice into a PSR (summing over Em, convolving with spectrum, and selecting non-empty CDS voxels). Caching the processed slices *after* these operations saves even more time and results in much less memory usage, since processing reduces the size of a response slice in memory by up to 1000-fold.

This patch therefore introduces the PSRCache class to cache processed response slices within FastTSMap. The cache's memory usage can be capped, though the small size of processed slices currently makes this limit unnecessary in practice. To improve performance should the cache be too small to hold every processed slice of a response, fitting now processes source directions in nested rather than ring order, as nested order results in greater locality for response slices. The plotting code has been modified to reflect this new ordering.

(4) Conversion from galactic-frame source directions to spacecraft-frame exposure maps currently creates and uses Astropy SkyCoord objects. Unfortunately, working with SkyCoords introduces significant overhead in the form of implicit calls to Astropy's coordinate frame transformation code. These calls happen even when no actual transformation is needed and were observed to constitute the majority of time spent in computing  exposure maps! Over hundreds to thousands of source directions, the cost to mapping becomes significant.

To remove SkyCoord-related overhead, FastTSMap has been modified to work primarily with 3-vector and raw colat/lon representations of source directions. Some of SpacecraftFile's APIs have been extended to accept these representations, and a new get_exposure() function was added to efficiently compute a sparse exposure map without using SkyCoords or Histograms. The existing get_dwell_map() function now uses get_exposure() internally.

(5) As a result of all the above changes, we no longer utilize Python's multiprocessing to parallelize mapping.  Parallelism is instead achieved at a lower level in fast_norm_fit (via Numba's prange construct) and in convolution/matrix multiply operations done on response slices. This design avoids the complexities incurred by relying on Python multiprocessing; in particular, it removes the overhead of launching multiple worker processes and avoids the difficulties of sharing response slices and other data structures between them. 

While it would still be nice to have a shared-memory multithreaded parallel implementation along the same lines as the original code, such a design will require advances in Python's free-threading support.

### API Changes

This patch makes some changes to simplify the ts_map APIs.  Both FastTSMap and MOCTSMap now expose a single fit() method, rather than differently named functions to do the actual fitting.  It is no longer necessary to externally generate a list of hypothesis coordinates for fitting; this is done internally based on an **nside** value provided to FastTSMap or a corresponding **max_nside** value for MOCTSMap.  And the plot_ts() method now consistently takes the result of fit() as an input, rather than sometimes storing this result in the class.

For MOCTSMap, the strategy used to decide which pixels to refine after each round of fitting can now be passed as an argument to fit().  MOCTSMap defines a Strategy object that implements the refinement rule. Using this API, it is now possible to select at runtime whether to refine, e.g., only the k highest-scoring pixels or all pixels above a threshold score, and whether to pad the set of refined pixels to also include neighbors of those selected by one of the above rules.

To support mapping with a galactic-frame response, this patch introduces a GalacticResponse class that shares its basic API with FullDetectorResponse, to the point that they can be used almost interchangeably in mapping. At some point, GalacticResponse and FullDetectorResponse should become subclasses of a common response class.

The above API changes are reflected in the ts_map tutorial notebook.

This patch also implements a long-pending change to exposure map computation in SpaceCraftFile: get_dwell_map() now takes a HealpixBase object to define the HEALPix grid for the map, instead of needing to open a response file each time it is called to determine the grid.  Call sites throughout cosipy have been updated to support this modified API, which allows the callers to avoid repeatedly opening the same response file.

### Performance Measurements

Here are some numbers, taken on one of my servers, that compare the ts_map code in the develop branch ("old") to that in the PR ("new"):

* First example in the ts_map tutorial notebook (nside=16, local-frame response):

  + Old on 36 cores: 133 s
  + Old on 8 cores:  354 s
  + New on 8 cores:  4.08 s

* Same example as above, but using MOCTSMap with max_nside=16:

  + Old on 36 cores: 12.9 s
  + Old on 8 cores:  13.4 s
  + New on 8 cores:  1.06 s

* Third example in the ts_map tutorial notebook (nside=16, galactic-frame response):

  + Old on 36 cores: 126 s
  + Old on 8 cores:  352 s
  + New on 8 cores:  3.22 s

### Potential for Future Extensions

The future of ts_map will likely involve substantial changes to how likelihood maps are computed. Response slices may be obtained by calling an ML model instead of reading a file. Compton events may be provided as a list rather than being precompiled into a CDS Histogram. And different PSRs may be applied to events occurring in different time/spacecraft attitude bins, rather than averaging all PSRs together as is done now.

The revised ts_map code can support these future innovations. Reworking fast_norm_fit to accept a tuple of PSRs and a corresponding tuple of event lists should be straightforward, and the CDS sparsity that we exploit now becomes even more important once events are divided into 1s time bins corresponding to different spacecraft attitudes. Also, the PSRCache can cache response slices computed by an expensive ML model to reduce the number of calls to the model during fitting.

Long-term, this design is not necessarily the best for fully unbinned likelihood mapping, as it does assume that the response/background models and time/spacecraft attitude are discretized in some way. However, for applications needing rapid mapping to support alert generation, such as the SDOC pipeline, it is not clear whether fully unbinned analysis will be fast enough. For performance reasons, it may be preferable to retain some degree of discretization, even if we no longer store the response itself in an HDF5 file. 

Another direction for extension is to add earth occultation and livetime support to exposure maps, matching the support currently offered by (more expensive to compute) scatt_maps. This is fairly easy to do and indeed is done in my ICRC work, but it requires putting some additional thought into reworking SpaceCraftFile's API.

### A Note on Test Coverage

Nearly all the code added by this PR is exercised by cosipy's test suite. In particular, the fast_norm_fit code is covered, with the exception of a couple of error paths in solve(). Unfortunately, the coverage tools cannot see calls to Numba JIT-compiled code, so they falsely claim that the majority of functions in fast_norm_fit.py are not covered. Coverage can be verified by observing that solve() includes calls to these functions or to other JIT-compiled functions that in turn call them.


Yong2Sheng:
Hi @jdbuhler — thanks for opening this pull request! I tested the updated code and tutorial notebooks with several sources; it worked smoothly, and the speed-up is really impressive!

I only have a few comments below. Please let me know if you think any of these are worth addressing, or if they’re good/trivial for merge.

## `GalacticResponse.py`

1. Minor nit: `h5py` is imported again at [line 50](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/response/GalacticResponse.py#L50) (looks like it’s already imported at the module level).

2. Overflow-bin support: [line 78](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/response/GalacticResponse.py#L78) indicates old response files with overflow bins are supported, and `get_counts()` explicitly handles that case. Should [`to_dr`](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/response/GalacticResponse.py#L185) and [`__getitem__`](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/response/GalacticResponse.py#L158C9-L158C20) also handle overflow-bin files (or alternatively, should we document that they’re not intended for that case)?

3. Minor nit: `Quantity` and `HealpixAxis` are imported but seem unused.

## `SpacecraftFile.py`

1. `_cart_to_polar()` docstring says it supports inputs with shape `[3]` or `[N, 3]`, but indexing at [lines 274–275](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/spacecraftfile/SpacecraftFile.py#L274-L275) seems like it would error for a `[3]` input.

2. `_cart_to_polar()` returns plain NumPy arrays, but `get_scatt_map()` calls [`r_colat.value` and `r_lon.value`](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/spacecraftfile/SpacecraftFile.py#L777). I think this would raise an `AttributeError` unless those are `Quantity`s. However, I tested the `get_scatt_map()` function, it didn't throw me an error. 

## `moc_ts_fit.py`

1. `plot_ts()` appears to ignore the user-provided `containment` at [line 299](https://github.com/cositools/cosipy/blob/994a32eb9fd1996da7775c984d8bfcf4a4adce09/cosipy/ts_map/moc_ts_fit.py#L299) (hardcoded to `0.9`).

## DetectorResponse.ipynb
1. It doesn't show me the figures in the notebook. I have to use `%matplotlib inline`.

## A general question
1. Does the response used in the tutorials include earth occultation?

## Unit test
I got this warning when running unit test on test_fast_ts_map.py, test_fast_norm_fit.py, test_spacecraftfile.py, test_galactic_response.py, test_full_detector_response.py, test_detector_response.py.
```bash
(/home/sheng2/Applications/conda_envs/cosipy_ts_acc) sheng2@node1624$ pytest tests/ts_map/test_fast_ts_map.py --cov=cosipy --cov-report term --cov-report html:tests/coverage_report
=============================================================================================================== test session starts ================================================================================================================
platform linux -- Python 3.10.19, pytest-9.0.2, pluggy-1.6.0
rootdir: /home/sheng2/Applications/conda_envs/cosipy_ts_acc/cosipy
plugins: cov-7.0.0
collected 3 items                                                                                                                                                                                                                                  

tests/ts_map/test_fast_ts_map.py ...
WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\m'


WARNING DeprecationWarning: invalid escape sequence '\s'
```
@israelmcmc I think we've seen this before. Could you please remind us how this was solved?

Thank you!

jdbuhler:
Thanks Yong! I addressed most of the issues you raised. A couple of comments:

* With respect to supporting the old, padded response Histogram, I don't want to add that support more pervasively in GalacticResponse; rather, we should ensure by DC4 that it is no longer required even in get_counts() by removing padding from any galactic responses that we offer through wasabi.  This is open issue #445.

* _cart_to_polar() will propagate units to the output if its input has units, which it happens to have in the case you noticed.  I changed the call site to remove the units from the input before calling the function.

* I'm not sure what the correct fix is to the notebook. I am honestly a little confused as to why Jupyter behaves as it does with %matplotlib inline.  It seems like adding this line somewhere is needed, but only the first time in the notebook that a plot is produced?

* The escape sequence warnings should go away if you just run the test suite again.  They are a side-effect of compiling the Python source (I believe in some dependency of the test case file, rather than in the file itself).  So far as I know, they are harmless.

Yong2Sheng:
Hi @jdbuhler , thank you for addressing the comments!

To fix the figure issue, we just add `%matplotlib inline` at the end of the importing part of the notebook. This happens to some other tutorial notebooks such as [Image Deconvolution notebook.](https://github.com/cositools/cosipy/blob/develop/docs/tutorials/image_deconvolution/511keV/GalacticCDS/511keV-Galactic-ImageDeconvolution.ipynb).

jdbuhler:
Got it -- I cleaned up the matplotlib inline directive as you suggested, and the TS_map notebook is now correctly showing all plots for me.

Yong2Sheng:
Thanks! I will merge the PR.

Yong2Sheng (approved):
Merge PR447

Issue #435, opened 2025-10-29
Ask to the RspConverter to take the default value of the norm if not found in the rsp file


GallegoSav:
For now RspConverter is setting the default norm to linear if nothing is found in the rsp file.
It should instead take the norm declared in the init (which is `Linear` by default) .

jdbuhler:
PR #448 has been created to address this problem.

jdbuhler:
Should this issue be closed now that we merged #448?

PR #448, opened 2025-12-02
Fix default norm behavior of .rsp file converter

jdbuhler:
Issue #435 observed that RspConverter was not properly setting the default spectral norm if an .rsp file was missing the SP header.  In fact, there are several issues around the default norm behavior, including an API (inherited from the previous response conversion code in the old FullDetectorResponse) that cannot actually accept arguments for all possible default norm types.

This patch fixes the default norm behavior of the RspConverter.  The constructor now takes arguments "norm" (a string, the norm name) and "norm_params" (a list) that are used only if the .rsp file has an empty or missing SP header.  norm parameters are uniformly validated and converted to the correct types, whether they occur in the file or as defaults, by a new validation function.  If the default norm is used, a synthetic SP header is generated so that conversion from HDF5 back to .rsp preserves the setting.

Along the way, I increased coverage of the eff_area calculation code by creating test cases with different default norms, fixing one bug.  I don't have a .rsp file that would let me test the default behavior with a Gaussian norm, as that norm is mostly unimplemented.

The API changes are backwards-compatible assuming that no default norm is provided to the constructor.  They deliberately break compatibility when a default is provided to ensure that the default arguments will be passed correctly.

It is perhaps worth considering if the default norm should be provided to the RspConverter constructor, or whether it should instead be an argument to the actual conversion function.


cosipy/response/RspConverter.py
```diff
@@ -341,10 +341,86 @@ def _read_response_header(self, rsp_file):
...
+        the detector.)  If the detector generates Compton events for
+        n_c of incident photons, then the effective area is
+
+          A(s, E_i) = a_i n_c(s, E_i)/n_i(s, E_i).
+
+        Because we store raw (integer) event counts, the count matrix
+        already includes the factor n_c. Hence, the area *correction*
+        A'(s, E_i) is simply a_i / n_i(s, E_i).
+
+        We assume that the simulation used to construct the response
```
GallegoSav, line 412: This is the current case of the simulation where we assumed isotropic illumination. However with the work of Pascal, looking at the effective area, it might change in the future in order to optimize the simulation input spectrum and spatial distribution. To have data distributed in the CDS more uniformly . 
What do you think @israelmcmc ? Should we open an issue to remember that ?

GallegoSav (approved):

PR #450, opened 2026-01-15
SourceInjector improvements

jdbuhler:
This patch makes a number of improvements to the SourceInjector class.  Most are minor correctness or efficiency fixes, but the most significant changes are

 * use the new GalacticResponse class to read galactic-frame responses, rather than digging into the underlying HDF5 file
 * add a test case for inject_model(), which didn't have one


Yong2Sheng (approved):
It looks good to me. Approve merge.

PR #451, opened 2026-01-16
Remove overflow bins from galactic response files

### LINKED ISSUE #445: galactic-frame response files still have track_overflow enabled

jdbuhler:
On Wasabi, the files in COSI-SMEX/DC2/Responses/PointsourceReponse [sic] are galactic-frame responses.  Although these files are Histograms, we do not read them using the regular Histogram.open() method; rather, we open the HDF5 file explicitly and read individual slices to create PSRs as needed, just as we do with spacecraft-frame responses.  This functionality is encapsulated in the GalacticResponse class.

Previously, most of these files were saved as Histograms with overflow bins, even though these bins are unused.  Checking for and accommodating the overflow bins required extra code in GalacticResponse.  This patch removes that code, instead rewriting all the galactic-frame responses used by the cosipy tutorials to remove them.  In particular, we have fixed up the following files:

511_gal_DC2_earthocc_nooverflow.h5
psr_gal_continuum_DC2_earthocc.zip
psr_gal_511_DC2.h5.gz
psr_gal_continuum_DC2.h5.zip

(The first of these already had the overflow stripped.) The other galactic-frame responses in DC2 are not referenced by any tutorial and have not been rewritten. _Please let me know_ if these other files need to be fixed up for use with the patched cosipy, or if they are no longer needed.  All future galactic responses created with histpy 2.0 or later will have overflow turned off by default and so will work fine.

The four fixed-up response files have been uploaded to Wasabi in the directory 

  COSI-SMEX/develop/Data/Responses/PointSourceResponse

(The extra "Data" matches the directory structure used by DC3 and DC4.)  The files have been renamed for consistency and repackaged for simplicity: the larger continuum responses are gzip'd, while the smaller 511 responses are left as raw .h5s because they are not particularly compressible.

The tutorial notebooks that use these four response files have been updated with to reflect their current names, checksums, and compression methods.  These include

image_deconvolution/511keV/GalacticCDS/511keV-Galactic-ImageDeconvolution.ipynb
source_injector/Extended_source_injector.ipynb
spectral_fits/extended_source_fit/diffuse_511_spectral_fit.ipynb
ts_map/TS_map_fitting.ipynb

All these notebooks have been verified to run successfully after the patch.  The Extended_source_injector notebook required more extensive changes, unrelated to the change in response file, to get it to run without error.

Note that the extended-source responses in COSI-SMEX/DC3/Data/Responses/extended_source_response have not been modified.  Although it would be useful to remove overflow bins from these responses as well, they are read using the ExtendedSourceResponse class, which loads the entire response into memory via histpy and so does not need to special-case for the presence or absence of overflow bins.  More generally, there are **many** files on Wasabi containing Histograms whose overflow bins could be stripped transparently to cosipy, and doing so would meaningfully reduce their storage and I/O costs.  But that is a separate project.

If accepted, this patch will close issue #445.


israelmcmc:
Thanks for working on this @jdbuhler.

> Please let me know if these other files need to be fixed up for use with the patched cosipy, or if they are no longer needed.
Yes, these are also needed. Despite not being part of the tutorials, they are required to complete other science tasks part of the data challenges (the tutorial just give you an example for a particular task). I'm tagging @ckarwin for awareness. This is independent of this PR, since these files are indeed not part of test of documentation.

Your changes look good to me and the tutorial are running fine for me as well. I'm merging.

jdbuhler:
Thanks.  I've uploaded overflow-free and consistently renamed versions of the other DC2 galactic-frame responses to the develop/Data/Responses/PointSourceResponses if you want to check.

PR #454, opened 2026-01-19
test_pipeline should use temporary output directory

jdbuhler:
test_pipeline runs two scripts that produce output files (bin.yaml and binned_data_local.hdf5).  Right now, these files are left in whichever directory the script was run from, which is to say wherever pytest was run.  This patch directs them to a temporary directory created by pytest instead so as not to litter the working tree.


GallegoSav:
Hi @jdbuhler , I see some commits from this pr are coming from your other [pull request ](https://github.com/cositools/cosipy/pull/457) on the DataIO. 

I suppose we need to merge first the pr 457 before this one

jdbuhler:
Ugh -- I thought I branched from develop.  Oh well.

GallegoSav (approved):

PR #455, opened 2026-01-20
Sorting the TimeTags of the unbinned file

GallegoSav:
@hirokiyoneda is using some time binning method for the image deconvolution which requires the TimeTags of the unbinned file to be sorted in ascending order.

Since we are merging multiple tra files during the mimrec extraction the TimeTags are shuffles when we use simulation.

So I added this sorting in the trafile converter method in the DataIO class. 

@ckarwin Could you review it please ? Should be quick to test.


ckarwin:
@GallegoSav Yes, I'll try to get to it tomorrow.

ckarwin (approved):
This is a small change. The logic is sounds and all tests have passed. It can be merged.

PR #456, opened 2026-01-20
Add support for FITS formatted orientation files

jdbuhler:
Orientation files are currently stored in a MEGALib-derived format, .ori, which is essentially a plain-text table that is read by NumPy.  This format is quite slow to read and will also not be acceptable in eventual conversations with HEASARC.  This patch implements storage of orientation files as an Astropy Table in FITS, which is both faster to read and closer to the eventual form needed for production use.

To support both .ori and FITS inputs, SpacecraftFile now has an open() class method that dispatches to one or the other format's reader depending on the file extension.  This method should be used everywhere instead of the .ori-specific "parse_from_file()" method (which should eventually be renamed and made internal).  The FITS reader and writer rely on Astropy's built-in gzip support to handle compressed FITS files transparently.

The actual FITS reading and writing code in SpacecraftFile is only a small part of the patch.  The majority of the patch is converting the rest of cosipy, including the test suite and notebooks, to use FITS instead of .ori files.  For the large orientation files that we download from wasabi, I've created FITS versions of them in the COSI-SMEX/develop subtree and converted the notebooks to use them.  (These files are not highly compressible, so I chose to use .fits instead of .fits.gz for them.)

Note on performance: while Astropy's FITS reader is much faster than the Numpy parser used for .ori format, the limiting factor in reading in an orientation file is now converting the coordinate axes in the file into SkyCoords and constructing the list of Attitudes, not the reading itself.


israelmcmc:
@jdbuhler Did you close it intentionally? Or was it due the handling of the branch names?

jdbuhler:
I screwed up my repo, and in fixing it, the branch went away.  I'll have to issue a new PR once I get things stabilized.  There was nothing wrong with the PR per se.

Jeremy

On Jan 23, 2026 09:23, Israel Martinez ***@***.***> wrote:
[https://avatars.githubusercontent.com/u/34419239?s=20&v=4]israelmcmc left a comment (cositools/cosipy#456)<https://github.com/cositools/cosipy/pull/456#issuecomment-3790483455>

@jdbuhler<https://github.com/jdbuhler> Did you close it intentionally? Or was it due the handling of the branch names?

PR #457, opened 2026-01-20
dataIO cleanup with faster .tra parsing

jdbuhler:
This patch does some cleanup in the dataIO directory, specifically the handling of unbinned and binned data files.  The main change is an overhaul of the unbinned .tra parsing code for greater speed, through removing some unneeded operations and vectorizing others.  There should be no difference in the results of the parse, and the new code is similar to the old code in terms of resilience to .tra format errors (that is, not very).

I observed about a 4x speedup in parsing the .tra file in the dataIO tutorial notebook.  Some of that comes from my changes, but some also comes from patching the scoords library, which is used in the code to construct an Attitude from X and Z axes.  @israelmcmc has a pending change to make this operation much faster, and I submitted an even faster version; this change should be in the forthcoming 0.0.3 release.

An intentional behavior change is this release is that cosipy will now throw an exception when given a .tra file with no X and Z pointing info, unless the user also provides an orientation file to supply the missing pointings.  Previously, the parsing code would set the X and Z axes to the same vector in this case, yielding a degenerate Attitude object whose construction fails with SciPy 1.15 and later.

Besides the changes to .tra parsing, the patch does a number of minor cleanups and small performance improvements to binned and unbinned data handling.


GallegoSav:
Hi @jdbuhler , does this patch require a different version of  `scoords` ? If yes, could you also update the version requirement in `setup.py` within this pr 

We just merged a [pr](https://github.com/cositools/cosipy/pull/455) which add a sort of the TimeTags in the tra_reader function . Could you update your branch with that to solve the conflict ?

jdbuhler:
The updated scoords (0.0.3+) is useful for performance, but the code will work with the current version (0.0.2), just more slowly (i.e., closer to the speed it ran pre-patch).  So I don't think we need a hard dependency on the new version.

I synced with latest develop, so should be fine with Hiroki's changes.

ckarwin:
@jdbuhler What if somebody wants to parse a tra file with no pointing info? It's common for people to run sims without ori files. That should be allowed.

ckarwin:
At least, that's why that option was there in the first place.

jdbuhler:
@israelmcmc suggested that this was not a well-defined use case, so I should remove it.  If we want to support this, what values should we put for the x and z axes?  We can't just set them both to (0,0), because the resulting Attitude is undefined.

GallegoSav:
> @israelmcmc suggested that this was not a well-defined use case, so I should remove it. If we want to support this, what values should we put for the x and z axes? We can't just set them both to (0,0), because the resulting Attitude is undefined.

There could be an option to find the attitude in the SpacecraftOrientation with the corresponding timetag. But you would need to pass the orientation file for that. So why not just use it in the first place during the simulation ? 

This should be mandatory, I did not know people were doing simulation without orientation file.


Anyway I just tried and indeed it is very fast ! I read 2.4G (PrimaryProtons)  in ~2min  :+1:

ckarwin:
I can't recall now, but at some point somebody needed this, which is why it was added. I guess you don't need an ori file if you only want to work in detector space, e.g., if you want to characterize a detector property as a function of off-axis angle, without being concerned about the actual position of the source on the sky.

ckarwin:
I think it doesn't hurt to keep the option. The use-case I mentioned above seems to be pretty well defined. Maybe it should be removed if it slows things down significantly, but I don't think that is the case (correct me if I'm wrong). I would point Z at the Galactic center and X orthogonal to that for a random direction.

israelmcmc:
The motivation to make it crash when the tra file has no pointing information and the user doesn't provide an ori file is to make it less error prone. In general, having the code choosing default valid values is bad practice because it's hard to note errors.

I think that a compromise is to let the user set a flag intentionally to avoid the crash, something like `no_attitude_ok = True`. Or, alternative, let the user set a constant attitude though the parameter that passes the file, like `sc_orientation = Attiutude(...`.

ckarwin:
Yeah, that's a good point. Maybe the second option is best, since it gives the user more freedom to define the constant attitude.

israelmcmc:
Yeah, I agree, second option is better

jdbuhler:
To be clear, there is no parameter to the tra parser that passes an orientation file.  The only option is to use the orientation data in the yaml file passed to the UnbinnedData constructor.

So do we allow a default attitude to be passed to the tra parser, independent of what is in the yaml file?

israelmcmc:
Thanks for the clarification @jdbuhler. Yes, this will require an additional parameter that overrides what's in the yaml file. 

This is related to, but out of the scope of, a larger goal of making every class capable to be initialized directly without the need (in addition to) to a yaml file.

jdbuhler:
So I can add an option to pass a SpacecraftFile to the parser to supply orientation data. At that point, do we also need another form of override (an Attitude), or passing the SpacecraftFile sufficient?

israelmcmc:
I'm thinking the extra parameter can support either a path to the ori file, or the exact fixed Attitude object.

jdbuhler:
If the parameter can be an Attitude object, must it contain just a single pointing?  I don't think anything else makes sense, as Attitudes do not carry timestamp information, so there is no sensible way to interpret more than one pointing from the Attitude alone.

jdbuhler:
As requested, I've added the ability to pass a spacecraft orientation file or a _single_ attitude to .tra parsing via a new, optional sc_orientation parameter.  If provided, this parameter overrides any other pointing info (either in the .tra file or in the orientation file specified in the .yaml).

GallegoSav (approved):

israelmcmc:
Thanks @jdbuhler! And @GallegoSav and @ckarwin for reviewing!

PR #459, opened 2026-01-27
Add polarized source support to SourceInjector

krishnatejavedula:
Added support for polarized source injection in source_injector.py and updated the point source injector tutorial with examples.


Yong2Sheng (approved):
Looks good! Approve merge.

PR #464, opened 2026-01-30
Add support for FITS formatted orientation files

jdbuhler:
Orientation files are currently stored in a MEGALib-derived format, .ori, which is essentially a plain-text table that is read by NumPy. This format is quite slow to read and will also not be acceptable for HEASARC. This patch implements storage of orientation files as an Astropy Table in FITS, which is both faster to read and closer to the eventual form needed for production use.

To support both .ori and FITS inputs, SpacecraftFile now has an open() class method that dispatches to one or the other format's reader depending on the file extension. The old, .ori-specific "parse_from_file()" method has been renamed "_open_ori()" and is only called from open(). The FITS reader and writer rely on Astropy's built-in gzip support to handle compressed FITS files transparently.  Per HEASARC, I've added a "VERSION" card to the FITS headers for orientation files, of the form YYYYMMDD.

The actual FITS reading and writing code in SpacecraftFile is only a small part of the patch. The majority of the patch is converting the rest of cosipy, including the test suite and notebooks, to use FITS instead of .ori files. For the large orientation files that we download from wasabi, I've created FITS versions of them in the COSI-SMEX/develop subtree and converted the notebooks to use them. (These files are not highly compressible, so I chose to use .fits instead of .fits.gz for them.)

Note on performance: while Astropy's FITS reader is much faster than the Numpy parser used for .ori format, the limiting factor in reading in an orientation file is now converting the coordinate axes in the file into SkyCoords and constructing the list of Attitudes, not the reading itself.


GallegoSav:
@israelmcmc What about this pr ? Should we merge it for DC4 ?

I tested the crab fit tutorial and it worked well using the ori fits file

israelmcmc:
Thanks @GallegoSav. Yes, I'm OK merging this. It's backward compatible so it shouldn't break anything in the data challenges repo. I'm tagging @ckarwin for awareness anyway

GallegoSav (approved):

GallegoSav:
Indeed the ori version is also still working.

Let's go 
image

PR #470, opened 2026-02-06
Fix coordinate frame issues with SpacecraftFile, COSILike, and ts_map

jdbuhler:
We recently identified an issue where COSILike produced incorrect results for a point source whose location is specified in ICRS rather than galactic coordinates.  This is a particularly insidious instance of a more general issue, which is that functions that take a source location specified as a SkyCoord don't always check that the frame of the coordinate is what they expect (and transform it if needed).

This patch fixes the observed issue in three parts of cosipy: COSILike (spectral fitting), SpacecraftFile.get_target_in_sc_frame(), and ts_map.  I've added test cases for the first two to verify that changing a source's coordinate frame doesn't affect the output (at least, up to reasonable numerical precision).  I can't directly add a test case for ts_map, because enumerating a grid of source directions in (say) ICRS gives a different set of hypotheses than enumerating an equal-sized grid of them in galactic.  But I did verify visually that, if I load the orientation file in ICRS and then plot the map and the true source in ICRS, they align as expected.

As long as I have to touch COSILike, I went ahead and added my performance patches to that file for computing the likelihood and expectation.  I also changed the remaining use of ICRS in the diffuse_511_spectral_fit tutorial (Case 3) to galactic, to match the change that was already committed for Case 2.  Thankfully, after the fix to COSILike, this is now only a cosmetic change :-).

In the longer term, we should review all of cosipy to see where it assumes that the inertial coordinate system being used is always galactic, and either enforce this assumption at call time (which I understand will be a feature of the interface branch) or make the code able to handle other frames (my preference where possible).  Notably, we use a GalacticResponse in a few places that should probably be an InertialResponse that communicates its frame.  This would let us specify, e.g., extended source responses in whatever frame we want.

### DIFF
--- a/cosipy/threeml/COSILike.py
+++ b/cosipy/threeml/COSILike.py
@@ -106,13 +106,6 @@ class COSILike(PluginPrototype):
         self._data = prepare_binned(data, cds_order)
         self._bkg  = prepare_binned(bkg, cds_order)
 
-        # Precomputed image response for extended
-        # sources
-        if precomputed_psr_file is not None:
-            logger.info("... loading the pre-computed image response ...")
-            self.image_response = ExtendedSourceResponse.open(precomputed_psr_file)
-            logger.info("--> done")
-
         try:
             data_frame = data.axes["PsiChi"].coordsys.name
             bkg_frame  = bkg.axes["PsiChi"].coordsys.name
--- a/cosipy/threeml/COSILike.py
+++ b/cosipy/threeml/COSILike.py
@@ -1,7 +1,10 @@
 import numpy as np
 
-from threeML import PluginPrototype
+import astropy.units as u
+
 from astromodels import Parameter
+
+from threeML import PluginPrototype
 
 from cosipy.response import (
     FullDetectorResponse,
@@ -12,8 +15,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 class COSILike(PluginPrototype):
-    """
-    COSI 3ML plugin.
+    """COSI 3ML plugin.
 
     Parameters
     ----------
@@ -36,11 +38,11 @@ class COSILike(PluginPrototype):
     nuisance_param : astromodels.core.parameter.Parameter, optional
         Background parameter
     coordsys : str, optional
-        Coordinate system ('galactic' or 'spacecraftframe') to perform
-        fit in, which should match coordinate system of data and
-        background. This only needs to be specified if the binned data
-        and background do not have a coordinate system attached to
-        them
+        Coordinate system (name of an inertial frame, such as
+        'galactic', or 'spacecraftframe') to perform fit in, which
+        should match coordinate system of data and background. This
+        only needs to be specified if the binned data and background
+        do not have a coordinate system attached to them
     precomputed_psr_file : str, optional
         Full path to precomputed point source response in Galactic
         coordinates
@@ -48,13 +50,34 @@ class COSILike(PluginPrototype):
         Option to include Earth occultation in fit (default is True).
 
     """
-    def __init__(self, name, dr, data, bkg, sc_orientation, 
+    def __init__(self, name, dr, data, bkg, sc_orientation,
                  nuisance_param = None,
                  coordsys = None,
                  precomputed_psr_file = None,
                  earth_occ=True,
                  **kwargs):
-        
+
+        def prepare_binned(data, cds_order):
+            """Given a Histogram describing binned data/model, convert it to a
+            bare array.  Make sure it is dense, lacks units, and
+            contains only CDS axes in the order specified.
+
+            """
+
+            # we *must* project before densifying, as some
+            # data sets are too big to densify otherwise
+            if tuple(data.axes.labels) != tuple(cds_order):
+                data = data.project(cds_order)
+
+            data = data.todense() if data.is_sparse else data
+
+            data = data.contents
+
+            if isinstance(data, u.Quantity):
+                data = data.value
+
+            return data
+
         # create the hash for the nuisance parameters. We have none for now.
         self._nuisance_parameters = {}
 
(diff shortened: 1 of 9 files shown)


jdbuhler:
Note that the patch coverage is pretty bad because we don't have a test case for COSILike that exercises the extended source functionality.  But this is well-exercised by the spectral fitting tutorial notebooks.

jdbuhler:
Image of ts_map output for first tutorial example using ICRS for hypothesis coordinates:

image

The plotting code assumes galactic, which is annoying to generalize because we cannot (I think) easily provide an arbitrary frame to the mollview.  For this test, I hacked the plotter to use 'C' instead of 'G' for the mollview and extracted the ra/dec of the true source instead of l/b. (Ignore the l 0 b 0 point and the "galactic" label)

cosipy/threeml/COSILike.py
```diff
@@ -63,30 +86,33 @@ def __init__(self, name, dr, data, bkg, sc_orientation,
 
         # User inputs needed to compute the likelihood
         self._name = name
-
+        self._dr = FullDetectorResponse.open(dr)
```
GallegoSav, line 89: Hi @jdbuhler , could you make the opening of the `FullDetectorResponse` also optional like the precomputed one ? Because we don't need it for an extended source fitting and we loose time to load it for nothing. 

There could be a check to see if ` model.point_sources` or ` model.extended_sources` are empty and just load accordingly the full response or extended source response(or both)

cosipy/threeml/COSILike.py
```diff
@@ -63,30 +86,33 @@ def __init__(self, name, dr, data, bkg, sc_orientation,
...
+            logger.info("--> done")
+
+        cds_order = tuple(self._dr.axes.labels[-3:])
+        if not all(ax in ("Em", "Phi", "PsiChi") for ax in cds_order):
+            raise ValueError("Response CDS axes must be Em/Phi/PsiChi")
+
+        self._data = prepare_binned(data, cds_order)
+        self._bkg  = prepare_binned(bkg, cds_order)
+
         # Precomputed image response for extended
```
GallegoSav: This should be remove no ?

cosipy/threeml/COSILike.py
jdbuhler: Let me look into this - I feel like the FDR gets used in some other way that makes it difficult to remove, but if I can remove it, sure. 

This isn't actually a big performance issue, as loading the FDR doesn't read the whole file into memory.  The much bigger cost in creating a COSILike is projecting the big sparse data and bg arrays.  Projecting a large sparse Histogram is very expensive, and the time is all in the guts of the Sparse library. But we can't just make these arrays dense because doing so blows out of memory.  If there were a method to load/bin them without the time axis, that might help.

(To put this in perspective, it takes way longer to do the projection than to actually run the optimizer in most of the notebooks. And I'm pretty sure the binning itself is much cheaper then the projection.)

jdbuhler: Ok, to respond to your initial request, the FDR gets used in the constructor to determine the order of the CDS axes.  But this can be done using the extended response as well if the FDR is not available.  Will just need a bit of hacking.

jdbuhler: I looked at the request about making the FDR optional.  It's easy enough to do internally to COSILike, but we really should come up with better names for the "dr" and "precomputed_psr_file" parameters.  The former is a FullDetectorResponse and is used only for point sources, while the latter is an ExtendedResponse and is used only for extended sources.  Maybe "response_for_point" and "response_for_extended"?  And we should be passing in the response objects themselves, not the filenames, so that repeated instantiations of the COSILike don't have to load a large extended source response each time.

I can certainly change the API, but it will require touching all the spectral fitting notebooks.  So I'm not sure these changes shouldn't be a separate PR, given that one unnecessary open of the FDR is not a serious performance issue right now. @israelmcmc , I'm not sure how this interacts with the interfaces branch.  What do you think?

jdbuhler: [Code changes to make the FDR optional without changing the external API are ready to be checked in and are tested against spectral fitting notebooks, but I would like @israelmcmc to weigh in first.]

israelmcmc: If it requires a significant API change, I'd just wait to merge the interfaces branch. In particular, I already worked out the issue you described @jdbuhler :
- We are passing around the FDR object, not the path. e.g. [here](https://github.com/israelmcmc/cosipy/blob/e019541adf225e02e124d4146ecada1200eba254/cosipy/response/instrument_response.py#L23).
- COSILike no longer exists, and it was divided into smaller classes performing specific tasks. 
   - The equivalent of passing `dr` is handled by the `BinnedInstrumentResponse` (the link above)
   - The equivalent of using `precomputed_psr_file` will be handled by `BinnedThreeMLExtendedSourceResponse` in @GallegoSav [PR](https://github.com/israelmcmc/cosipy/pull/11/changes).
   
Currently we only use the pre-computed PSR for extended source, but we can also use it for point sources. Similarly, adding this feature would be way easier with the interfaces refactoring.

@jdbuhler The easiest way to see how the interfaces refactoring changes thing is to follow this [script](https://github.com/israelmcmc/cosipy/blob/interfaces/docs/api/interfaces/examples/crab/example_crab_fit_threeml_plugin_interfaces.py), which is the equivalent of the binned Crab fit tutorial. The only reason I haven't asked people to start using the interfaces branch instead of develop is because the unit test are broken, and I know that makes it hard to work with. I'm getting close to fixing them though, hopefully I'll finish this week and I'll ping you.

jdbuhler: Ok, then I will not check in the changes in response to @GallegoSav 's comment and will wait for the interfaces branch.

I would like to see the stuff already in this PR go in, *unless* you have a solution in interfaces to avoid the transform_to() calls I added in COSILike and get_target_in_sc_frame.  I don't love those because they create new SkyCoords, which is an expensive operation.  If you have a different place to check the frames and convert only if needed, I will just change those two lines to assertions.  The other stuff in the PR should still be useful as-is.

cosipy/threeml/COSILike.py
GallegoSav: @jdbuhler Could you remove this duplicate of the precomputed_psr_file loading so I can merge this pr ?

cosipy/threeml/COSILike.py
israelmcmc: I think the equivalent of these transformation occur here in the interfaces branch:
https://github.com/israelmcmc/cosipy/blob/e019541adf225e02e124d4146ecada1200eba254/cosipy/response/instrument_response.py#L176-L185

This function rotates the response (i.e. the differential effective energy), which is in local coordinates, for a given source `direction` into a specific frame defined by `axes` (the data axes). I'm also using `transform_to()`, so I'm not sure if it is faster. However, @jdbuhler, I used your previous improvements using your `rotate_coords` for the rotation of each PsiChi direction (or at least that was my intention, I wonder if I reverted some of your optimization by accident during the refactoring).

jdbuhler: The fact that you are transforming to ICRS internally rather than the frame of the source does suggest that you didn't pick up my changes to response rotation.  I specifically remember changing this so that, if everything is already in galactic, we don't have to go to ICRS and back.

Really, I suspect we need to be protecting any performance-relevant transform_to() calls with a test to see if the frame is already what we want.  transform_to() always creates a new SkyCoord, even if the frame does not change, and that creation process seems to be very costly.  Part of my changes to COSILike were to avoid creating a SkyCoord on every iteration of optimization, which added at least 10% to the cost of minimization in the GRB spectral fitting notebook.

israelmcmc: I see, yeah. I'll fix this instance, but in general we'll have to profile the code again after the interfaces refactoring. Surely there are more instances like this one, unfortunately.

GallegoSav:
let's go 
image

GallegoSav (approved):

israelmcmc:
Thanks @GallegoSav and @jdbuhler!

PR #472, opened 2026-02-09
Add the fluctuate option to the source injector

### LINKED ISSUE #471: Poisson fluctuate SourceInjector outout
Currently the SourceInjector returns the mean expected number of counts. In order to obtain a more realistic simulated dataset, it would be good to have a flag that user can set to Poisson fluctuate the output. This mean that the number of counts on each bin will be a random sample (an integer) coming from a Poisson distribution whose mean is the current output (the mean expectation). Each output will be different from the pervious one allowing to test our statistical method and run Monte Carlo simulation. 

It should be something simple like 

```python
if self._fluctuate:
    injected[:] = np.random. Poisson(injected)

return injected
```

GallegoSav:
I added this option if the user wants to have poisson fluctuation each time the source injector is running. 

This is important for fitting and MC simulations.

#471

### DIFF
--- a/cosipy/source_injector/source_injector.py
+++ b/cosipy/source_injector/source_injector.py
@@ -288,7 +288,7 @@ class SourceInjector():
                      data_save_path=None,
                      project_axes=None,
                      polarization=None,
-                     fluctuate=False):
+                     fluctuate=True):
         """
         Build an injected source by combining all the sources in a
         model.  Each injected source is stored by name in a dictionary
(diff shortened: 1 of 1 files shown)


krishnatejavedula (approved):
I’ve reviewed the PR, and the changes look good. I approve merging.

GallegoSav:
I set the fluctuate option to True per default.

israelmcmc:
Thanks @GallegoSav and @krishnatejavedula!

PR #476, opened 2026-02-12
Speedup for SpacecraftFile math

jdbuhler:
This patch is principally a performance improvement for SpacecraftFile.  The earth occultation math has been pulled out into its own function (where it will eventually be available to use for DwellMaps after the interfaces branch goes in) and rewritten to reduce its cost.  It also has the ability to cache some source-independent intermediates to speed up repeated earth occultation math for different sources.  This is helpful if, for example, we want to consider earth occultation in the ts_map code.  The code for  scatt_map creation is also sped up by reusing some of the functions I wrote for faster exposure map computation.

As a little aside, this patch also enhances the GalacticResponse class I introduced in a previous patch to optionally take and honor a datatype for response slices, matching what the FullDetectorResponse can do  now.  This is useful if we want to carry out an analysis in single precision for performance.

_Important note_: this patch bumps cosipy's minimum required version of scipy to 1.15.  The reason for the bump is that scipy prior to 1.15 throws errors on a vector Rotation of length 0, and therefore on a vector Attitude of length 0 (which can happen if, for example, the source is occluded at every time point).  FullDetectorResponse's point source response code had an ugly workaround that I introduced for this issue, which I have now removed, but it also comes up in the modified scatt_map creation code.

I don't believe we have any issues in current cosipy or histpy that would prevent us from using the newer scipy.  The self tests and notebooks, to the best of my knowledge, work fine with this version. Currently, using Python 3.10 prevents the dependencies from pulling in the latest scipy (1.17), but in my testing with Python 3.12, even that version will be fine provided that my recent patch to mhealpy is released.


GallegoSav:
@israelmcmc Is this change on the scatt map computation will interfere with your interfaces branch and the way how you use the livetime ?

israelmcmc:
@GallegoSav  yes, a little, but I'm already in contact @jdbuhler and he is aware of the new code. We agreed to fix this once things are settled. This can be merged if there's no issue otherwise.

GallegoSav (approved):

jdbuhler:
I'll need to deal with the new representation of earth pointings in interfaces. But I didn't see anything particularly new in livetime representation, except for some new external APIs from SpacecraftHistory. Did I miss something?

israelmcmc:
@jdbuhler In the interfaces branch, I currently create a histpy.Histogram whose contents are the livetime and edges are the timestamps from telemetry. This is hidden under the livetime property, so we can change the underlying code if it turnes out it's too slow.

jdbuhler:
I have not played with the TimeAxis stuff in histpy, but if I can pull the raw numpy array out of the livetimes, and the livetime binning still matches the Attitudes stored in the SpacecraftHistory, it shouldn't be a big issue.  If the time axis no longer lines up with the Attitudes, I'll have to do some more work.

israelmcmc:
@jdbuhler One possible problem I can foresee is that TimeAxis uses astropy's Time underlaying representation with two doubles in order to obtain sub-nanosecond absolute time resolution. I wanted to keep this general in histpy since some experiment do need it, but in cosipy we don't really, since the SC does not provide that level of precision anyway. If the current implementation is too slow, we can refactor to use another representation internally in SpacecraftHistory

Issue #477, opened 2026-02-12
Test response-data axes mismatch


israelmcmc:
The axes of the current test response [test_full_detector_response.h5](https://github.com/cositools/cosipy/blob/develop/cosipy/test_data/test_full_detector_response.h5) do not match the corresponding axes of the data [test_spectral_fit_data.h5](https://github.com/cositools/cosipy/blob/develop/cosipy/test_data/test_spectral_fit_data.h5) and background  [test_spectral_fit_background.h5](https://github.com/cositools/cosipy/blob/develop/cosipy/test_data/test_spectral_fit_background.h5).

I inspected the elements in this test:
https://github.com/cositools/cosipy/blob/c557fc2001469954eddf447c425534a9575dde45/tests/threeml/test_spectral_fitting.py#L12-L18

```
$ FullDetectorResponse.open(dr).axes['Em'].edges
<Quantity [  100.   ,   158.489,   251.189,   398.107,   630.957,
            1000.   ,  1584.89 ,  2511.89 ,  3981.07 ,  6309.57 ,
           10000.   ] keV>

$ background.binned_data.axes['Em'].edges
<Quantity [ 150.,  220.,  325.,  480.,  520.,  765., 1120., 1650., 2350.,
           3450., 5000.] keV>

$ data.binned_data.axes['Em'].edges
<Quantity [ 150.,  220.,  325.,  480.,  520.,  765., 1120., 1650., 2350.,
           3450., 5000.] keV>
```

I realized this while fixing the units tests in the `interfaces` branch, which does check that the axes match. In `develop` only the number of bins matter, which is why it went unnoticed. 

I'll fix in `interfaces` (by hacking the edge values, so I can recover the same numbers), but I'm opening this here since it affect `develop` as well.

israelmcmc:
Closed by #369

PR #480, opened 2026-02-13
Make test_point_source_spectral_fit a circular test

### LINKED ISSUE #357: spectral fitting test in test suite is unstable with low-resolution response
I have been trying to replace the o1 continuum response used in the cosipy test suite (test_data/test_full_detector_response).  In doing so, I am seeing very unstable behavior from the spectral fitting code -- it sometimes crashes inside threeML,  Specifically, I get the following threeML warning:

WARNING   100.0 percent of samples have been thrown away because they failed the constraints on the parameters. This results might not be  suitable for error propagation. Enlarge the boundaries until you loose less than 1 percent of the samples. (    analysis_results.py:1739)

and it then crashes at random_variates.py:202, where it calls np.percentile on what I think is an empty array.

I can avoid the crash by increasing the upper limit on the background_cosi parameter from 5 to 50 (it is still fit at a value < 3), but I still see warnings about discarding 99% of samples.

I suspect that I'm asking the fitting code to do something impossible -- the response may just be too coarse to generate a result that looks anything like the observed CDS data.  Israel suggested I open an issue about this in hopes that someone can come up with a small test case for spectral fitting that behaves sensibly.

israelmcmc:
Create fake test data and background using the same response as the fit. This is circular because if test the fitting routine itself, not the simulations or anything else.

Fixes #357 (because the response, despite being coarse, is the true response as far as the algorithm cares)  and #477 (because the energy bins edges are by definition the same).

This was the last unit test that was still failing in the interfaces refactoring, but I believe it's due to unrelated reasons (those in the issues.


israelmcmc:
Btw, I also changed Band to PowerLaw. Even with a better response, people have reported that it is really difficult to fit the Crab to a Band function due to the narrow COSI energy resolution (the lower energy index is not well constrained). We should do the same with the tutorials.

eneights (approved):
This looks good to me, thanks @israelmcmc!

israelmcmc:
Thanks @eneights

PR #481, opened 2026-02-16
close subfigures after plotting

jdbuhler:
Certain plotting functions in cosipy create subplots via pyplot.subplots() or similar.  Each such call generates a new figure to hold the subplots, whose lifetime does not exceed that of the calling function.  If these figures are not closed after use, they may stay around indefinitely, triggering a warning from matplotlib about "too many open figures" and presumably consuming memory until they are garbage-collected (if that is even possible).  I encountered this issue while generating thousands of likelihood map plots for GRBs with the ts_map code; the warning kicks in after producing 20 such plots over a few seconds.

This patch ensures that the figures created by functions using subplot are explicitly closed.  If the function might be called to display a plot in a notebook, it is necessary to call plt.show() *after* any savefig() call (to avoid disrupting the saved figure) but *before* plt.close(); otherwise, the figure is never displayed.

I patched the ts_map code and one subplot use in the XSpec support code.  I also put an explicit close() into the plotting function in the pipeline code, but that function had plt.show() commented out, so I did not re-enable it (though I did move it after the savefig() call).

I don't believe this is an issue for plotting uses that do not generate a new figure each time; in any case, many such uses in cosipy already have a correspnding close() call.

### DIFF
--- a/cosipy/pipeline/src/plotting.py
+++ b/cosipy/pipeline/src/plotting.py
@@ -133,6 +133,7 @@ def plot_fit(sou, cts_exp, figname):
     ax[1].set_ylabel("Obs-Model/Err")
     ax[1].set_xlabel("Energy (keV)")
     #
-    # plt.show()
     plt.savefig(figname)
+    # plt.show()
+    plt.close(fig)
     return ()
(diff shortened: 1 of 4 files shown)


GallegoSav (approved):

GallegoSav:
image

israelmcmc:
Thanks @jdbuhler  and @GallegoSav. This is certainly an improvement, so it was good that @GallegoSav merged it. However, the root issue is our less-than-ideal use of matplotlib. I explain this in #482. That's for later though.

PR #483, opened 2026-02-16
Temporary workaround for 3ML circular import while waiting for release.

israelmcmc:
Installing from source, where the pkg_resources deprecation error and the logging circular import error have been fixed.

Will remove with the next 3ML release


jdbuhler:
Should we remove the pin of matplotlib at the same time?  I believe it was there originally because of colormap issues in the old threeML that have since been fixed in dev, and I don't think we have any problems with newer matplotlib in cosipy itself.

(I think old matplotlib pins us to old numpy, which is also undesirable and unnecessary for cosipy

israelmcmc:
@jdbuhler Yeah, I'll make the change

israelmcmc:
Self-merging as maintenance task...

PR #486, opened 2026-02-16
Fix breakage caused by recent polarization-related merge

jdbuhler:
A recent change to COSILike to support polarization info now calls SpacecraftFile.get_dwell_map() with a polarization argument.  There is no support for such in SpacecraftFile.  This patch restores the original call to get_dwell_map().

We also fix one line in the GRB spectral fitting notebook with a broken print call around a 3ML display call.

Ideally, the test suite would have caught this, but to do so, we'd need a COSILike test case that uses spacecraft-local rather than inertial coordinates, which requires some new data files.  This is related to the fact that I'm getting complaints about this patch being all uncovered lines of code. (It passes the test suite, but the "fail" status is based on the low coverage)

### DIFF
--- a/cosipy/threeml/COSILike.py
+++ b/cosipy/threeml/COSILike.py
@@ -10,7 +10,12 @@ from cosipy.response import (
     FullDetectorResponse,
     ExtendedSourceResponse
 )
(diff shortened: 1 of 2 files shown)


israelmcmc:
Thanks @jdbuhler . Makes sense. I opened issue #487  about the missing test and I'm merging this.

PR #488, opened 2026-02-16
Add ML polarization notebook to auto-runnable tutorials

jdbuhler:
This patch adds the new maximum likelihood polarization tutorial to the list runnable from tutorials.yml.  I just run both it and the ASAD tutorial together, rather than creating separate entries for each.

I had to fix one issue with the tutorial notebook where there was some stuff commented out that should not have been.  I also did some very minor cosmetic cleanup and am checking in the expected output.

### DIFF
--- a/docs/tutorials/polarization/maximum_likelihood_method.ipynb
+++ b/docs/tutorials/polarization/maximum_likelihood_method.ipynb
@@ -2,6 +2,7 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "2bcdf955",
    "metadata": {},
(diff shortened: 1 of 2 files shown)


israelmcmc:
Thanks @jdbuhler. Would it be possible to have each one on it's own entry, please? I'd like to remember which one can be done in parallel. The other ones where I have multiple notebooks per entry is because the second one depends on the outputs of the first one.

jdbuhler:
Done

GallegoSav (approved):

israelmcmc:

PR #490, opened 2026-02-17
Get polarization Stokes tutorial working in tree

jdbuhler:
The recently introduced Stokes polarization tutorial needed updating to use the FITS orientation API and to use the correctly named response file.  Once I got it working, I added it to tutorials.yml .

I'm resolutely resisting the urge to tinker with the Stokes code for now, though it would benefit substantially from vectorization.


israelmcmc:
Thanks @jdbuhler!

PR #491, opened 2026-02-20
Fix minor error in polarization_stokes.py

eneights:
The format that the azimuthal angles were returned in was incorrect, and @nmik wrote this fix. This issue caused the output of the example notebook to be incorrect, and I updated that as well.


israelmcmc:
Thanks @eneights . I confirm that I get the same result. Merging!

PR #494, opened 2026-02-25
Back-compatibility patch for updated FITS orientation files in interfaces

jdbuhler:
The FITS orientation files created by the interfaces branch add units to every column; this was the subject of the recent PR #19 to interfaces.  This patch ensures that develop can use these new FITS files without trouble by having open() strip units from quantities that did not have them in the older FITS files.

The bulk of the patch updates the tutorials with the correct md5 checksums for the new versions of the FITS orientation files in the develop subtree of wasabi.


jdbuhler:
Withdrawing because interfaces has been merged to develop

Issue #413, opened 2025-09-30
Tool to split the data for phase-resolved analysis


israelmcmc:
Given an ephemeris --e.g. start time and period-- split the event data based on their phase.

israelmcmc:
Related to #238. 

After the interfaces refactoring, this can be an EventSelectionInterface implementation.

israelmcmc:
Closed by #497

Issue #344, opened 2025-04-10
DataIO refactoring


israelmcmc:
This issue superseeds #249 and #338, closed by #342. 

This is meant to start the discussion on how DataIO will look like longer term. None of this is critical, it's only refactoring, and therefore it's low priotirity

I'll start with my thoughts. I propose:
- [ ] `UnbinnedData` and `BinnedData` should both derive from `DataIO`. `DataIO` should contain all common functions and properties (currently `UnbinnedData` derived from `DataIO`, and `BinnedData` from `UnbinnedData`)
- [ ] Both `UnbinnedData` and `BinnedData` should have the methods `open()`, `write()` and `plot()`.  For `BinnedData`, these will mostly call histpy methods, instead of duplicating the IO code (tagging @jdbuhler  since he had thoughts about this)
- [ ] Methods like `get_raw_spectrum` and `get_raw_lightcurve` should return another `BinnedData` object (with a reduced number of axes). They'll now how to `open()`, `write()` and `plot()` themselves already. Similarly for other axes --e.g. psichi, phi.
- [ ] `BinnedData` should have a classmethod `from_unbinned()` which takes an `UnbinnedData` object and bins it.
- [ ] `UnbinnedData` should have a method called `to_binned()` which internally calls `BinnedData.from_unbinned()`. This is a convenience method, but doesn't duplicate code.
- [ ] `config` file should become optional, only if the user finds them convenient. There should be a way to use `DataIO` programatically without config files.
- [x] (Done in #369 )~We should create an e.g. `EventSelection` class. This will take an event (or event list), and output `True` of `False` (or an array that works as a mask) depending on whether an event passed a cut or not. Derived classes from `EventSelection` can implement different types of cuts --e.g. energy, time, if it's in the SAA, if it's a GTI, number of hits, hit location, etc. Multiple event `EventSelection` objects can be concatenated. The `UnbinnedData` object should be able to take these as an input to filter out events. This mechanism should replace specific methods like `select_data_energy`, `select_data_time`, `filter_good_data`, `cut_SAA_events`, etc. 
- [ ] The binning information should come from a histpy `Axes`, which can in turn from from a response.

israelmcmc:
I forgot to click the button. They are closed now, thanks for pointing it out.

israelmcmc:
When we fix this, fix also the light curve tutorial. See
https://github.com/israelmcmc/cosipy/pull/22

jdbuhler:
[...]

PR #497, opened 2026-02-27
Add Phase-Resolved Analysis Tutorial and Tools.

### LINKED ISSUE #395: Tool to obtain a phasogram
This is an intermediate task toward a phase resolved analysis.

As a step, we need a tool that takes unbinned data and an ephemeris, and results in a histpy Histogram with the phases as a axis.

Using only source counts for now is fine. Eventually, in a later step, we'll need event selections, taking into account the background, and various timing corrections.

KoothodilAbhijithAugustine:
# Description

Resolves #395 and #413

This draft PR introduces a complete tutorial and a set of supporting Python tools for performing **phase-resolved pulsar analysis** within the `cosipy` framework. 

Currently, the workflow demonstrates how to process Data Challenge 2 (DC2) FITS data for the Crab pulsar and its corresponding albedo background. The end goal of this pipeline is to prepare isolated on-pulse and off-pulse FITS files that are ready for downstream COSI Spectral Fitting.

### Key Additions
* **Phase-Resolved Analysis Tutorial (`example_notebook.ipynb`)**: A step-by-step Jupyter Notebook guiding users through combining unbinned data, slicing by Mission Elapsed Time (MET), assigning phases, plotting the pulse profile, and selecting phase intervals.
* **`phase_assigner.py`**: A new module that reads a pulsar ephemeris file (e.g., `crab.par`) and calculates/appends the exact rotational phase to individual photon events in a FITS file.
* **`plot_pulse_profile.py`**: A plotting utility to generate and visualize the folded light curve/pulse profile from the phase-assigned data.
* **`phase_selector.py`**: A filtering tool to extract specific phase windows (e.g., on-pulse and off-pulse regions) and save them as new FITS files.


cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,147 @@
...
+            batch = list(itertools.islice(it, n))
+            if not batch: return
+            yield batch
+
+class PhaseSelector(EventSelectorInterface):
+    """
+    Selects events based on pulsar phase. 
+    Highly optimized for FITS files via NumPy vectorization.
+    """
+    def __init__(self, ephemeris_file, pstart, pstop, batch_size=10000):
```
hiyoneda: Better to allow users to put a list of phase interval ranges.

cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,147 @@
...
+        self.ephemeris_file = ephemeris_file 
+        self.pstart = float(pstart)
+        self.pstop = float(pstop)
+        self._batch_size = batch_size
+
+    def _get_vectorized_mask(self, phases: np.ndarray) -> np.ndarray:
+        """Core logic applied across an entire NumPy array instantly."""
+        pstop_norm = self.pstop % 1.0 if self.pstop > 1.0 else self.pstop
```
hiyoneda: If the multiple ranges are allowered, we can constrain the pstop and pstat to be from 0 to 1.

cosipy/phase_resolved_analysis/phase_assigner.py
```diff
@@ -0,0 +1,70 @@
...
+
+class PhaseAssigner:
+    """
+    Reads a pulsar ephemeris and assigns PULSE_PHASE based on MET directly.
+    """
+    def __init__(self, par_file):
+        self.params = self._parse_par_file(par_file)
+        
+        # Parse F0 (Frequency) or P0 (Period)
+        if 'F0' in self.params:
```
hiyoneda: For future work: let's put "F1" and "F2" in the future, not this time.

cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,147 @@
+import logging
+import numpy as np
+import itertools
+import os
+from astropy.io import fits
+
+logger = logging.getLogger(__name__)
+
+# --- ROBUST IMPORTS (Safety Switch) ---
+try:
```
hiyoneda: The new interface has been merged, so this part can be updated.

cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,147 @@
...
+            return self._get_vectorized_mask(events['PULSE_PHASE'])
+
+        # Fallback for generic iterables of objects
+        for chunk in itertools_batched(events, self._batch_size):
+            phases = np.array([getattr(e, 'pulse_phase', -1.0) for e in chunk])
+            mask = self._get_vectorized_mask(phases)
+            for sel in mask:
+                yield bool(sel)
+
+    def filter_events(self, events, output_fits=None, template_fits=None):
```
hiyoneda: ~~filter_event, _save_fits_fast, and save_fits can be removed because these functionalities should be implemented as a part of DataIO and are duplicated if they are added here.~~

cosipy/phase_resolved_analysis/plot_pulse_profile.py
```diff
@@ -0,0 +1,109 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from matplotlib.gridspec import GridSpec
+
+class PlotPulseProfile:
+    """
+    Generates a 3-panel figure: Pulse Profile, Phaseogram, and Significance Test.
+    Optimized for direct FITS table input (vectorized).
+    """
+    def __init__(self, data, n_bins=50, n_time_bins=50):
```
hiyoneda, line 18: I'd be happy if you could add a docstring here to explain what kind of "data" is assumed as an input of this class.

hiyoneda: Also, what kinds of plots will be generated

cosipy/phase_resolved_analysis/time_selector.py
```diff
@@ -0,0 +1,126 @@
...
+    class EventSelectorInterface:
+        def select(self, events): raise NotImplementedError
+    def itertools_batched(iterable, n):
+        it = iter(iterable)
+        while True:
+            batch = list(itertools.islice(it, n))
+            if not batch: return
+            yield batch
+
+class TimeSelector(EventSelectorInterface):
```
hiyoneda: you can delete this class.

hiyoneda (commented):
Thanks @KoothodilAbhijithAugustine, I am checking your codes. Let me share some comments.

hiyoneda:
I checked the notebooks and found some issues. This is a draft pull request, so you may already have noticed these points and plan to resolve them in the next pull request. Let me share them anyway.

- The first cell raises errors due to wrong importing. `from time_selector import TimeSelector` should be `from cosipy.phase_resolved_analysis import TimeSelector, PhaseSelector, PlotPulseProfile`
- `fetch_wasabi_file` is not imported but called. The same for `analysis.combine_unbinned_data`.
- For `fetch_wasabi_file`, you can add checksum values to avoid downloading the same data multiple times. The checksum value can be checked with `md5` command on mac.
    - MD5 (Crab_DC2_3months_unbinned_data.fits.gz) = dc261c4adf86a44e0349d47c72185a99
    - MD5 (albedo_photons_3months_unbinned_data.fits.gz) = 4c125410d8f127d7e12682f008d5651d
    - MD5 (20280301_3_month_with_orbital_info.ori) = 206122c81670bbd8765b43c9d4872bd4
- On the 5th cell, the filename extension is changed to '.fits' from '.fits.gz', which causes an error. Overall, it would be great to ensure the notebook runs without any modifications.
- To assigning pulse phase, PhaseAssigner should be added in `__init_.py`, and it can be imported as `from cosipy.phase_resolved_analysis import PhaseAssigner`.
- FileNotFoundError: [Errno 2] No such file or directory: 'crab_with_albedo_1d.fits'. It should be 'crab_with_albedo_1d.fits.gz'.

With modifications, the notebook worked well in my environment.

hiyoneda:
Actually, I got somewhat different results for the 2D phaseogram. Do you have any idea on this?

Screenshot 2026-03-25 at 12 08 21

hiyoneda:
@israelmcmc
Is there an example of how to perform event selection using the new interface with COSI data? We currently use the filter_events method in the PhaseSelector class for pulse-phase-based event selection, and would like to migrate to the new interface instead.
I am also trying to understand the overall data-handling structure under the new interface. The DataIO class provides the event list as a dictionary (cosi_dataset) as before, and the event selection class appears to support TimeTagEmCDSEventDataInSCFrameFromDC3Fits — but it does not seem to be used in the DataIO unbinned class. Are there plans to integrate DataIO with the new interface more?

israelmcmc:
@hirokiyoneda Here's how the event selection works, but please see my point at the end that it might now be the best way to go about this for the phase.

Currently the only example of an [`EventSelectorInterface`](https://github.com/cositools/cosipy/blob/develop/cosipy/interfaces/event_selection.py) is [`TimeSelector`](https://github.com/cositools/cosipy/blob/develop/cosipy/event_selection/time_selection.py), but it's enough to understand how it works.

The `EventSelectorInterface` is pretty simple: it receives a list of events of a certain type, and for each one it returns a boolean flag that signals whether it was selected or not. The event data type determines which parameters a given selector can use for the passing criteria. For example, the `EventSelectorInterface` uses `TimeTagEventDataInterface`, which is guaranteed to have `time`. You can pass a particular event data that has even more information, through inheritance, but `EventSelectorInterface` will only consider `time`. For example, `TimeSelector` supports the particular implementation that you mentioned, `TimeTagEmCDSEventDataInSCFrameFromDC3Fits`, because it inherits from `TimeTagEmCDSEventDataInSCFrameFromArrays` > `TimeTagEmCDSEventDataInSCFrameInterface` > `TimeTagEventDataInterface`. 

Note that `EventSelectorInterface` doesn't know where the data comes from. The `DataIO.UnbinnedData` class is the one responsible for handling the files on disk. Internally, `TimeTagEmCDSEventDataInSCFrameFromDC3Fits` uses it to load the events. Thats currently the extent of the integration between the interfaces and `DataIO`, but it can be improved! However, I think we'll need to work on #344 first (or do both at the same time).

Ideally, `UnbinnedData` should be able to wraps it's data using instances of `EventDataInterface`. It can have it's own selection methods for convenience, but it should internally delegate the calculations to a particular `EventSelectorInterface`. 

For the case at hand, phase selection, you would need:
1. Define the `EventWithPhaseInterface`/`EventWithPhaseInterface` (akin to `TimeTagEventInterface`/`TimeTagEventDataInterface`).
2. Define specific implementation of the above classes, which can actually be instatiated e.g. `EventWithPhase`/`EventDataWithPhase` (akin to `TimeTagEmCDSEventInSCFrame`/`TimeTagEmCDSEventDataInSCFrameFromArrays`).
3. Write a `PhaseSelector(EventSelectorInterface)` class, that can take a `EventWithPhaseInterface` an returns the list of bool flags.

*However*, using a `PhaseSelector` might not be the most efficient way to do phase analysis, since you'd have to loop through the same event list multiple time for each phase selection. It would be better to use the phase information to split the data directly, one set of data for each phase range. Or, if you want it binned, you can have a "phase" axis in your histogram and bin it directly.

israelmcmc:
On a separate topic, note that we'll also need a method to update the exposure in `SpacecraftHistory` based on the phase selection. This is how we can get the right normalization without changing the rest of the analysis.

hiyoneda:
Thanks @israelmcmc for the thorough explanation.

I understand that the `DataIO` refactoring is still ongoing. For now, I thought that it would be better to decouple `PhaseSelector` from `EventSelectorInterface` and make it a standalone class that directly filters a `DataIO.UnbinnedData` instance. This keeps things simple and lets us move forward with DC4 using Abhi's current implementation as a base. It might be possible to use `EventSelectorInterface` but at this moment we need to think about how to save the filtered data (something like opening `TimeTagEmCDSEventDataInSCFrameFromDC3Fits` with filtering -> converting `TimeTagEmCDSEventDataInSCFrameFromDC3Fits` to `UnbinnedData ` -> saving data as a fits could be possible but complicating). Once the `DataIO` refactoring is complete, we can revisit and properly integrate `PhaseSelector` into the interface hierarchy.

For `SpacecraftHistory` and the response, I think the same approach would be good; for now, we write a utility function that manually adjusts the exposure (and response if needed) based on the phase selection, and migrate to passing an `EventSelectorInterface` once the infrastructure is ready.

Does this two-step approach sound reasonable to you @israelmcmc @KoothodilAbhijithAugustine ?

israelmcmc:
@hiyoneda Yes, I'm OK with the plan for PhaseSelector/UnbinnedData. It is indeed hard to achieve a nice integration with the interfaces without refactoring DataIO first, so we can wait until then.

For `SpacecraftHistory`, that's independent of the event interfaces, since they don't have a concept of uptime (it's just a list of events). I'm thinking there should be an new interface e.g.

```python
class PhaseEphemeris(Protocol):

    def get_phase(time:Time) -> np.ndarray:
        """"
        Assign a phase between 0 an 1 to each timestamp.
        """"
        
    def get_duty_cycle(time_start:Time, time_stop:Time, phase_start:float, phase_stop:float) -> u.Quantity:
        """"
        Compute the time spend between the phase range for each time range.
        """"
```

`PhaseEphemeris` will then be inherited by classes that can open specific formats or can be initialized from specific user parameters (e.g. t0 + period).

Then, we can add a method to `SpacecraftHistory` that applies this duty cycle e.g.

```python
class SpacecraftHistory:

    def update_ephemeris(ephemeris: PhaseEphemeris, phase_start:float, phase_stop:float):
        """
        Updates live time based on a phase cut.
        """
        # Here you can call ephemeris.get_duty_cycle()
``` 

The advantages:
- The specific `PhaseEphemeris` will know the fastest way to compute the duty cycle
- `SpacecraftHistory` delegates this calculation, but it know what's the best way to update it's own live time.
- It'll be easy to add new ephemeris formats or models if needed.

cosipy/phase_resolved_analysis/phase_selector.py
hiyoneda: Since we decided to make this class independent of the event selection interface, we can leave these methods.

hiyoneda:
@israelmcmc Thanks. Adding the new protocol and the method to SpacecraftHistory in this PR sounds good.
@KoothodilAbhijithAugustine Can you work on this?

KoothodilAbhijithAugustine:
> Actually, I got somewhat different results for the 2D phaseogram. Do you have any idea on this?
> 
> Screenshot 2026-03-25 at 12 08 21

The results that you have on the panel are correct for 1 day (86400 s). The results that is on the notebook is of 10 days (10 * 86400s).

hiyoneda:
Can you move the unittest #549 to this PR? The unittest coverage is checked when we merge new code.

KoothodilAbhijithAugustine:
Hi @hiyoneda, all checks are now passing. This is ready for your final review!

cosipy/phase_resolved_analysis/phase_assigner.py
```diff
@@ -0,0 +1,105 @@
+import logging
+import numpy as np
+from astropy.io import fits
+
+logging.basicConfig(level=logging.INFO)
```
hiyoneda: please delete this line.

cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def _validate_intervals(self, intervals):
+        """Ensures all phase values are strictly within [0, 1]."""
+        if not isinstance(intervals, list):
+            intervals = [intervals]
+            
+        validated = []
+        for start, stop in intervals:
+            if not (0.0 <= start <= 1.0 and 0.0 <= stop <= 1.0):
+                raise ValueError(f"Phase boundaries must be between 0 and 1. Got: ({start}, {stop})")
```
hiyoneda, line 28: Can you check whether start < stop?

cosipy/phase_resolved_analysis/phase_selector.py
```diff
@@ -0,0 +1,80 @@
...
+        """Applies all interval ranges to a NumPy array using bitwise OR."""
+        combined_mask = np.zeros(phases.shape, dtype=bool)
+        
+        for pstart, pstop in self.intervals:
+            if pstart <= pstop:
+                # Normal range
+                mask = (phases >= pstart) & (phases <= pstop)
+            else:
+                # Wrap-around range (e.g., 0.9 to 0.1)
+                mask = (phases >= pstart) | (phases <= pstop)
```
hiyoneda: If start < stop is checked, this part can be simplified.

cosipy/phase_resolved_analysis/plot_pulse_profile.py
```diff
@@ -0,0 +1,148 @@
...
+        ax_phaseogram.set_title("Phaseogram")
+        plt.colorbar(im, ax=ax_phaseogram, label="Counts/bin")
+
+        # --- Panel 3: Significance (Bottom Left) ---
+        if len(t_elapsed) > 1:
+            sort_idx = np.argsort(t_elapsed)
+            sorted_phases = self.phases[sort_idx] * 2 * np.pi 
+            sorted_times = t_elapsed[sort_idx]
+            
+            # Cumulative Z^2_2 statistic (2 harmonics)
```
hiyoneda: Please add a comment that we make the maximum harmonic number a parameter in the future, like "#TODO: xxx".

tests/phase_resolved_analysis/test_phase_tools.py
```diff
@@ -0,0 +1,48 @@
...
+        
+        calculated_phases = (test_times * assigner.f0) % 1.0
+        np.testing.assert_allclose(calculated_phases, expected_phases, atol=1e-7)
+    finally:
+        if os.path.exists(par_path):
+            os.remove(par_path)
+
+def test_phase_selector_wrap_around():
+    """Check wrap-around logic (e.g., 0.9 to 0.1)."""
+    selector = PhaseSelector(intervals=[(0.8, 0.2)])
```
hiyoneda: This part needs to be modified if we constrain start < stop, as [(0.0, 0.2), (0.8, 1.0)]

hiyoneda (commented):
Thanks @KoothodilAbhijithAugustine 
I've checked the code and added some minor comments. Once they are resolved, I will approve the merge.

Just let me confirm this: do you plan to work on this as a separate PR?

> For SpacecraftHistory, that's independent of the event interfaces, since they don't have a concept of uptime (it's just a list of events). I'm thinking there should be an new interface e.g.

> class PhaseEphemeris(Protocol):

    def get_phase(time:Time) -> np.ndarray:
        """"
        Assign a phase between 0 an 1 to each timestamp.
        """"
        
    def get_duty_cycle(time_start:Time, time_stop:Time, phase_start:float, phase_stop:float) -> u.Quantity:
        """"
        Compute the time spend between the phase range for each time range.
        """"
> PhaseEphemeris will then be inherited by classes that can open specific formats or can be initialized from specific user parameters (e.g. t0 + period).

> Then, we can add a method to SpacecraftHistory that applies this duty cycle e.g. class SpacecraftHistory:

    def update_ephemeris(ephemeris: PhaseEphemeris, phase_start:float, phase_stop:float):
        """
        Updates live time based on a phase cut.
        """
        # Here you can call ephemeris.get_duty_cycle()

>

KoothodilAbhijithAugustine:
> Thanks @KoothodilAbhijithAugustine I've checked the code and added some minor comments. Once they are resolved, I will approve the merge.
> 
> Just let me confirm this: do you plan to work on this as a separate PR?
> 
> > For SpacecraftHistory, that's independent of the event interfaces, since they don't have a concept of uptime (it's just a list of events). I'm thinking there should be an new interface e.g.
> 
> > class PhaseEphemeris(Protocol):
> 
> ```
> def get_phase(time:Time) -> np.ndarray:
>     """"
>     Assign a phase between 0 an 1 to each timestamp.
>     """"
>     
> def get_duty_cycle(time_start:Time, time_stop:Time, phase_start:float, phase_stop:float) -> u.Quantity:
>     """"
>     Compute the time spend between the phase range for each time range.
>     """"
> ```
> 
> > PhaseEphemeris will then be inherited by classes that can open specific formats or can be initialized from specific user parameters (e.g. t0 + period).
> 
> > Then, we can add a method to SpacecraftHistory that applies this duty cycle e.g. class SpacecraftHistory:
> 
> ```
> def update_ephemeris(ephemeris: PhaseEphemeris, phase_start:float, phase_stop:float):
>     """
>     Updates live time based on a phase cut.
>     """
>     # Here you can call ephemeris.get_duty_cycle()
> ```
> 
> >

I'll work on them as a separate PR.

cosipy/phase_resolved_analysis/phase_selector.py
KoothodilAbhijithAugustine: Not yet.

cosipy/phase_resolved_analysis/plot_pulse_profile.py
KoothodilAbhijithAugustine: done

tests/phase_resolved_analysis/test_phase_tools.py
KoothodilAbhijithAugustine: OK

tests/phase_resolved_analysis/test_phase_tools.py
```diff
@@ -0,0 +1,49 @@
...
+        calculated_phases = (test_times * assigner.f0) % 1.0
+        np.testing.assert_allclose(calculated_phases, expected_phases, atol=1e-7)
+    finally:
+        if os.path.exists(par_path):
+            os.remove(par_path)
+
+def test_phase_selector_wrap_around():
+    """Check wrap-around logic (e.g., 0.8 to 0.2)."""
+    selector = PhaseSelector(intervals=[(0.8, 1.00)])
+    selector = PhaseSelector(intervals=[(1.00, 0.2)])
```
hiyoneda: I think that it should be `    selector = PhaseSelector(intervals=[(0.0, 0.2), (0.8, 1.0)])`?

KoothodilAbhijithAugustine: Thank you!

hiyoneda (approved):
Thanks, @KoothodilAbhijithAugustine, I approve it.

israelmcmc:
Thanks @KoothodilAbhijithAugustine and @hiyoneda !

PR #499, opened 2026-03-09
Add unpolarized unbinned source injector capability [Yong2Sheng's]

israelmcmc:
I'm bringing @Yong2Sheng from my fork to develop, now that the interfaces branch is merged.

## Summary

This PR introduces an initial implementation for **unpolarized** unbinned continuum point source injector in the spacecraft frame. It follows the existing line injector workflow, but with additional steps to (1) compute the integrated detected rate across an energy band and (2) sample detected photon energies across that band.

### 1) Define a continuum spectrum (differential photon flux)
A power-law spectrum is used as the working example in the tutorial:

$$S(E)=K\left(\frac{E}{\mathrm{piv}}\right)^{\mathrm{index}},
\qquad
[S(E)] = \frac{\mathrm{ph}}{\mathrm{cm^2}\times\mathrm{s}\times\mathrm{keV}}
$$

where $K$ is the normalization (differential flux at the pivot energy $\mathrm{piv}$).

### 2) Convolve the spectrum with effective area
The detected count-rate density is

$$
w(E) \equiv S(E)\times A_{\text{eff}}(E),
\qquad
[w(E)] = \frac{\mathrm{ph}}{\mathrm{s}\times\mathrm{keV}}.
$$

### 3) Integrate $w(E)$ to get the total detected rate and total counts
The integrated detected photon rate is

$$
I_0 = \int_{E_\mathrm{min}}^{E_\mathrm{max}} w(E) dE,
\qquad
[I_0] = \frac{\mathrm{ph}}{\mathrm{s}}.
$$

In practice, the spectrum might not be analytical. And the effective area is also usually not analytical as well. **So we do not perform the integral directly. Instead, we will use a very fine energy grid to obtain the integral by trapezoids.**

The expected detected counts is

$$
\mu = T\times I_0,
$$

and the realized detected counts are sampled as

$$
N \sim \mathrm{Poisson}(\mu).
$$

### 4) Build an inverse-CDF sampler to draw detected photon energies
- Construct a discrete CDF from $w(E)$ on the energy grid via cumulative trapezoidal integration, and normalized it by $I_0$ to $[0,1]$.
- Draw a uniform distribution from [0,1): $u \sim \mathrm{Unif}(0,1)$.
- Apply inverse transform sampling: $E_i = CDF^{-1}(u)$, where $CDF^{-1}$ is implemented using `scipy.interpolate.interp1d` with linear interpolation.

### 5) Simulate each sampled energy
Feed the sampled photon energies (and their multiplicities, although they should show up only once) into `UnpolarizedIdealComptonIRF.random_events(...)` to produce the event list.

## Implementation notes
1. A new class `RandomEventDataFromContinuumInSCFrame` is responsible for continuum injection.
2. For performance and clarity, I cache the following intermediates with `@cached_property`:
    - `energy_grid` 
    - `unpolarized_aeff`
    - `unpol_counts`
    - `incident_photons`
3. A structured return container for detected-count (`unpol_counts`) bookkeeping:
    - `UnpolCountsSummary(rate_density, rate, total_expected_counts)`

## What is not included yet
- Polarization handling for the continuum branch (this PR is unpolarized only).


Yong2Sheng:
I will add the unit tests after the changes have been reviewed and approved. I’d like to avoid adding them too early, since they may need to be updated if the code changes during review.

israelmcmc:
See my comment in #528
> This is also related to https://github.com/cositools/cosipy/pull/499: if the user passes a UnbinnedThreeMLModelFoldingInterface, we can inject event data right from the same class.

It's fine for now, but eventually this capability should be moved to the SourceInjector class.

Yong2Sheng:
Hi @ldigesu , I have added the saving function. Could you please take a look? Thank you!

ldigesu:
Hello Yong,
The tutorial runs ok for me; it simulates and saves the output as a fits file correctly.  My main comments are about explanations in the tutorial and sanity checks.  

I suggest adding an introductory paragraph at the beginning of the notebook, describing what this simulator does and what it is expected to produce as outputs. Something like the summary written in this PR should be ok.

Secondly, I thinks the tutorial is missing some sanity checks of the outputs. All the plots run correctly to show the simulated data, but there is no comparison with what was input, thus no checks that what was produced by the simulator is indeed what we expected. The saved fits output cannot be used in a binned analysis: the binning function does not work, because it can't find the Time Tags of the photons. Thus, my suggestion as a sanity check is the following: instead of showing as an example a random point source, you could take one of the point sources of DC3 e.g., a bright GRB or the Crab nebula, and simulate exactly the same position, spectrum, source duration, ecc.. Then you can produce plots from your simulator and from the fits file produced by Megalib and show that they are consistent. Let me know if this could work.

Thirdly, I remind you Israel's comment to PR #528 (here above), that is relevant to this PR.

Finally, a last minor comment: I find the last plot showing the distribution of the photons in the CDS not particularly explanatory.  I understand that there are no Galactic coordinates in these data, thus no way to visualize a ring around the expected source position. However, maybe you could try to you apply some cuts in Energy or Phi angles before plotting and see if point clusters. I suggest in any case to add a note for inexperienced COSI users explaining that in local/spacecraft coordinates, no ring/cone can be visualized.

Thanks! Let me know If everything is clear,
Laura

israelmcmc:
Thanks for the comments @ldigesu 

About this:
> Finally, a last minor comment: I find the last plot showing the distribution of the photons in the CDS not particularly explanatory. I understand that there are no Galactic coordinates in these data, thus no way to visualize a ring around the expected source position. However, maybe you could try to you apply some cuts in Energy or Phi angles before plotting and see if point clusters. I suggest in any case to add a note for inexperienced COSI users explaining that in local/spacecraft coordinates, no ring/cone can be visualized.

Maybe we can do a better job for that plot and the explanation, but the point of the plot is that if you take all the event with a given phi (color coded) then they should make a circle with radius=phi around the location of the source (the cross). It's a visualization of the Compton cone, using a colorbar for Phi instead ofa "z-axis".

PR #500, opened 2026-03-09
Line background estimation [hirokiyoneda's]

israelmcmc:
I'm moving @hirokiyoneda's PR from my fork to the main repo
https://github.com/israelmcmc/cosipy/pull/31

It adds spatial filtering (on the PsiChi plane) options to `generate_bkg_model_histogram` to mitigate Poisson fluctuations in the background model on the PsiChi plane.

Three new parameters are added to `generate_bkg_model_histogram`:
- `smoothing_fwhm`: applies Gaussian smoothing to the PsiChi map
- `l_cut`: applies a spherical harmonic low-pass filter, retaining only multipole moments up to `l_cut` (features on angular scales larger than ~180/l_cut degrees). Negative values introduced by the filter are clipped to zero.
- `rebin_phi`: merges adjacent Phi bins before applying the spatial filter, then restores the original binning. The spatial pattern of each Phi bin is replaced by the filtered template, while the total count of each original Phi bin is preserved.

Note: `smoothing_fwhm` and `l_cut` cannot be specified simultaneously and will raise a `ValueError`.

### DIFF
--- a/cosipy/background_estimation/LineBackgroundEstimation.py
+++ b/cosipy/background_estimation/LineBackgroundEstimation.py
@@ -2,12 +2,14 @@ import logging
 logger = logging.getLogger(__name__)
 
 from histpy import Histogram, Axis, Axes
+import healpy as hp
 import astropy.units as u
 import numpy as np
 import matplotlib.pyplot as plt
 from scipy.optimize import curve_fit
 from scipy import integrate
 from iminuit import Minuit
+from itertools import product
 
 class LineBackgroundEstimation:
     """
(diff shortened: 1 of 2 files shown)


israelmcmc:
Closing and reopening to see if the codecov/project reference commit is updated.

israelmcmc:
Thanks you @hirokiyoneda. Looks good to me and I can confirm the notebook runs for me as well. I'm merging:

Note: I think we should remove this when we revisit the documenation.
> We may apply smoothing to the background histogram, which may help to mitigate Poisson fluctuation in the model.

Issue #453, opened 2026-01-19
Removing the source+bkg merging step


GallegoSav:
It is an extra step which can take time when you have an heavy file and not necessary at all. 
And also it ducplicates the number of needed files for tutorial. Instead people could just download bkg and signal file and define the `data` as `signal.binned_data.project('Em', 'Phi', 'PsiChi')+bkg.binned_data.project('Em', 'Phi', 'PsiChi')`. 

Like for the crab : 

```
cosi = COSILike("cosi",                                                        # COSI 3ML plugin
                 dr = dr,                                                      # detector response
                 data = crab.binned_data.project('Em', 'Phi', 'PsiChi')+bkg.binned_data.project('Em', 'Phi', 'PsiChi'),   # data (source+background)
                 bkg = bkg.binned_data.project('Em', 'Phi', 'PsiChi'),         # background model 
                 sc_orientation = sc_orientation,                              # spacecraft orientation
                 nuisance_param = bkg_par,                                     # background parameter
                 earth_occ = True)                                             # Option to account for Earth occultation
```

This would require updating all the tutorials and removing this extra step and all the signal+bkg combined file though

PR #501, opened 2026-03-09
Removed signal+bkg step from GRB Tutorial [jbotha-sketch's]

israelmcmc:
I'm moving @jbotha-sketch's PR from my fork to the main repo. See https://github.com/israelmcmc/cosipy/pull/30

Removed the download and use of grb_bkg since it can be constructed from the combination of the grb signal and the bkg data. This fix is equivalent to the solution for issue #453 which was recently merged to develop on the main repository.


israelmcmc:
Thanks @jbotha-sketch! Looks good to me and it runs well. I'm merging

PR #514, opened 2026-03-12
Add tests for MacOS with M chip

israelmcmc:
This add the macos-latest runner for the units tests, which uses and M chip:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners

This is motivated by the different backends pytorch can use, but it's a good idea in general to check multiple systems.


israelmcmc:
Maintenance task. Self-merging.

PR #515, opened 2026-03-13
Point Source Spectral Fitting with Normalizing Flows Approximation for Response and Background

### LINKED ISSUE #389: Add NN instrument response to cosipy
This involves @scipascal moving his code to cosipy.

scipascal:
## Description
This pull request introduces a neural-network-powered response and background approximation to `cosipy`, which powers the unbinned Poisson likelihood analysis of continuum point sources. For more information, refer to the tutorial notebook `example_crab_fit_normalizing_flows.ipynb` or my thesis, _"Development of an Efficient Response Description for the COSI MeV 𝜸-Ray Telescope."_

### Hardware Testing
The implementation has been verified on the following environment:
* **OS:** Debian GNU/Linux 12 (Bookworm)
* **CPU:** 2 × AMD EPYC 9454 (96 physical cores total)
* **Memory:** 1.1 TiB
* **GPU:** 4 × NVIDIA RTX A5000 (24 GiB VRAM each)

> [!IMPORTANT]
> Multiprocessing or PyTorch may encounter issues on AMD GPUs or M-Processors.

### Installation
Running this code requires additional packages. Install them using the `ml` extra:
`pip install -e ".[ml]"` (includes `torch`, `sphericart[torch]` and `normflows`)

Additionally, you’ll need two checkpoint files to run `NFResponse` and `NFBackground`, which you can find on Wasabi (thanks @GallegoSav).

Closes #389


cosipy/background_estimation/nf_unbinned_background.py
```diff
@@ -0,0 +1,118 @@
...
+from cosipy.interfaces.event import EventInterface
+from cosipy.interfaces.data_interface import TimeTagEmCDSEventDataInSCFrameInterface
+from cosipy.data_io.EmCDSUnbinnedData import TimeTagEmCDSEventInSCFrameInterface
+from cosipy.interfaces.background_interface import BackgroundDensityInterface
+from cosipy.util.iterables import asarray
+
+from importlib.util import find_spec
+
+if find_spec("torch") is None:
+    raise RuntimeError("Install cosipy with [ml] optional package to use this feature.")
```
GallegoSav: put the pip command to use explicitly. `pip install -e . "[ml]"`

GallegoSav:
@israelmcmc I tried the tutorial notebook with the brightest GRB ( [GRB_bn090424592](https://github.com/cositools/cosi-sim/tree/main/cosi_sim/Source_Library/DC3/sources/GRBs/bn090424592) ) using GPU or CPU. 

Both worked well.
CPU is slower but still affordable ~20-30 min for precomputation (which can be saved ).

israelmcmc:
Thanks @scipascal and @GallegoSav. I'll try it out.

@GallegoSav for that GRB that lasted 20-30 min in a CPU, so you know how many event you were analyzing? 

Or in general, I want to get a sense of the cost per event in both GPUs and CPU, maybe @scipascal knows too.

scipascal:
> Or in general, I want to get a sense of the cost per event in both GPUs and CPU, maybe @scipascal knows too.

@israelmcmc The benchmark that I performed in my thesis was 34e4 evaluations per second for the probability density function on four NVIDIA RTX A5000. Lets assume you need 60 evaluations per event for numerical integration. Then you get 5.7e3 events per second. Thats very slow and therefore we need the caching and optimized integration schemes to use it.

israelmcmc:
Thanks @scipascal. Do you have the same performance numbers of a CPU?

scipascal:
@israelmcmc, for CPU only (see the hardware specifications) I get around 225 events per second compared to the 5.7e3 events per second with 4 GPUs.

israelmcmc:
Thanks for checking @scipascal!

israelmcmc:
Heads up: it's likely that the unit tests of this PR will have the same problems I described here: https://github.com/cositools/cosipy/pull/519#issuecomment-4100578292

israelmcmc:
@scipascal I added your tutorial to run_tutorials.yml and fixed some file path so I can run it automatically:
https://github.com/israelmcmc/cosipy/commit/24acf9261a1672c10b1d552e142b8fd8e396c149

with e.g.

```
~/software/cosipy/docs/tutorials/run_tutorials.py ~/software/cosipy/docs/tutorials/run_tutorials.yml -o tutorials --wasabi_mirror ~/cosi/data/wasabi/cosi-pipelines-public/ --tutorial spectral_fit_normalizing_flows
```

I haven't managed to run it though, can you please check if it runs for you automatically with the changes and command above? And if so, can you merge them to your branch please?

I keep seeing a "Kernel died" error, but I think it's due to memory limits or a bad setup, not something in your code. I'll keel checking, but if you've seen this before let me know.

GallegoSav:
@israelmcmc Could you try this example ? This is using CPU and try to fit `GRB_bn090424592`   . 
[crab_unbinned_tutorial-CPU.ipynb](https://github.com/user-attachments/files/26301983/crab_unbinned_tutorial-CPU.ipynb)

To see if you have still the "kernel died" issue.

You will have to replace `FasterTimeSelector` by `TimeSelector` in the cell which make the time selection. We implemented locally a faster time selector , to correct for the issue opened by Pascal. Though I think it has been corrected by Jeremy since ?

israelmcmc:
@GallegoSav Unfortunately the kernel is still dying for me for the GRB example. It happens in this cell:
```
psr = UnbinnedThreeMLPointSourceResponseIRFAdaptive(data, irf, sc_orientation,force_energy_node_caching=True)
psr.set_integration_parameters(batch_size = 1000) # Increasing the batch_size will increase the RAM consumption
```

I tried lowering the batch size to see if that fixed the issue.

I still thing it might be the memory, just maybe in another step.

Can you post your `FasterTimeSelector` somewhere, pelase? The current one is indeed pretty slow, and having a faster iteration should help debugging this.

GallegoSav:
@israelmcmc here the time selector 

```
class FasterTimeSelector(EventSelectorInterface):

    def __init__(self, tstart:Time = None, tstop:Time = None):
        """
        Assumes events are time-ordered
        
        Selects events that fall within ANY of the time intervals defined by
        corresponding pairs of (tstart, tstop).

        Valid combinations:
        - (None, None): No time constraints
        - (Scalar, None): Single lower bound only
        - (None, Scalar): Single upper bound only  
        - (Scalar, Scalar): Single time interval
        - (List, List): Multiple time intervals (same length required)

        Parameters
        ----------
        tstart: Time, scalar Time, or None
            Start time(s). If list, tstop must also be a list of same length.
        tstop: Time, scalar Time, or None
            Stop time(s). If list, tstart must also be a list of same length.
        batch_size: int, default 10000
            Number of events to process at once
        """
        if tstart is not None and tstop is not None:
            if not tstart.isscalar == tstop.isscalar:
                logger.error("tstart and tstop must both be scalar or both be list.")
                raise ValueError

        elif tstart is None and tstop is not None:
            if tstop.isscalar == False:
                logger.error("When tstart is None, tstop must not be a list.")
                raise ValueError

        elif tstart is not None and tstop is None:
            if tstart.isscalar == False:
                logger.error("When tstop is None, tstart must not be a list.")
                raise ValueError
        
        # tstart is None and tstop is None -> OK.
        
        # Convert scalars to lists for uniform processing
        if tstart is not None and tstart.isscalar == True:
            tstart = Time([tstart])

        if tstop is not None and tstop.isscalar == True:
            tstop = Time([tstop])
        
        # length check
        if tstart is not None and tstop is not None:
            if len(tstart) != len(tstop):
                logger.error(f"tstart and tstop must have same length.")
                raise ValueError

        self._tstart_list = tstart
        self._tstop_list = tstop

    def _select(self, events:TimeTagEventDataInterface, early_stop:bool = False) -> Iterable[bool]:
        event_times = events.time 

        if self._tstart_list is None and self._tstop_list is None:
            return np.ones(len(event_times), dtype=bool)

        if self._tstart_list is None:
            result = event_times <= self._tstop_list[0]

        elif self._tstop_list is None:
            result = event_times > self._tstart_list[0]

        else:
            result = np.zeros(len(event_times), dtype=bool)
            for i in range(len(self._tstop_list)):
                result |= (event_times > self._tstart_list[i]) & (event_times < self._tstop_list[i])
        #if early_stop and self._tstop_list is not None and len(event_times) > 0:
        #    if event_times[-1] > self._tstop_list[-1]:
        #        stop_idx = np.searchsorted(event_times, self._tstop_list[-1], side='right')
        #        return result[:stop_idx]

        return result
```

israelmcmc:
Thanks @GallegoSav . I'll give it another try

ckarwin:
@israelmcmc, @scipascal, @GallegoSav  I have Pascal's branch installed. I want to test his PR using the automatic tutorial pipeline. Israel, I see from the comments above that you implemented this already, but I don't think Pascal has merged this into his branch. I think this should be tested before the PR can be merged, and so I prefer to wait till it's updated. Or do you see another reason I should test this now, and go through the notebook manually?

israelmcmc:
@ckarwin You can go ahead and run it from my branch automatically. I didn't change anything in @scipascal's code, just the notebook setup up, so you would be testing the same underlaying code.

ckarwin:
@israelmcmc Ok, I see that you only changed 2 files. I copied them over and I'm running now.

ckarwin:
@israelmcmc It's still running, but to clarify, I'm running this on the palmetto cluster (not my Mac).

ckarwin:
@israelmcmc @GallegoSav The tutorial has been running on the cluster now for 15 hrs, using CPUs. How long is it expected to take?

How do I get the tutorial to use GPUs? Does it detect the availability of GPUs automatically? I tried submitting the job to a node with a GPU but it still ran with just CPUs.

israelmcmc:
It's not urgent right now, we need to first double check that everything works, but please see the discussion around here https://github.com/cositools/cosipy/pull/519#issuecomment-4165594015 about how we plan to handle classes that need the `[ml]` dependencies. 

In summary:
- Importing pytorch comes with some difficulties, and I'd like to isolate all classes that need it. 
- These classes should go into `.ml` submodules e.g. `cosipy.response.ml`
- The `ml/__init__.py` files should check if cosipy was installed with `[ml]` through the flag `cosipy.with_ml`. 
- Nothing from there should be imported by default with cosipy --i.e. it must not be picked up by the root `cosipy/__init__.py`. 

Currently, our unit test only work if healpy and pytorch are installed from conda (not from pip). pip's healpy and pytorch work well independently, but they clash when imported simultaneously. This is explained in the [installations instructions](
https://github.com/cositools/cosipy/blob/develop/docs/install.rst)

scipascal:
@israelmcmc, I’ve merged your changes. Running the tutorial automatically works for me. If you haven’t checked, could you and @ckarwin make sure you’ve set the devices you want to use in the notebook (CPU or the specific CUDA devices)? If you think it makes more sense I can also change the default device to CPU.

israelmcmc:
I was able to run @GallegoSav's GRB example using CPU only on a macbook with an M4 chip. It took 30 min. I run [this version](https://gist.github.com/israelmcmc/158cb38f0b40e328285e9d31f3142fe8) (I had to make some small changes:  paths, update a parameter's name, and add the FasterTimeSelection) on commit 77265f01d58bef0dfea898a53103cee3270cb301. I'm not sure what changed compares to the last time that was giving me the "kernel died" error, except for the fact that I'm not using healpy+pytorch from conda. See issue #548. I don't see why #548 matters in this case though, so it might be something else.

Here's my suggestion on how to proceed with this PR:
- For now, let's change the default in the tutorial to CPU. I think that's enough to prove that the code works.
- Let's change the tutorial to a GRB. We can eve use a dimmer GRB than the one used by @GallegoSav (GRB_bn090424592) and still get a good fit. Or, alternatively, select a shorter time window within GRB_bn090424592. The goal of the tutorial is only to explain to the user how to use the code, so the lighter the better.
- In parallel, we can keep trying reproduce you results with GPUs (I haven't managed to do this successfully). I'm OK merging this PR as long as it works on CPU. Once we understand better how to make the GPU backend work with other system, we can simply add it to the documentation.

What do you think @scipascal? 

Also, can you please add unit tests? We will also need those before merging this PR.

Side note: I used the FasterTimeSelector that @GallegoSav posted above. It's indeed much faster! I'm changing the regular TimeSelector in #543 so it no longer be the bottleneck

scipascal:
@israelmcmc I updated the tutorial.
- Correct imports from `ml` folders
- GRB bn090424592 instead of the Crab
- By default CPU and no compilation
- The background file is already selected
- Slightly more comments

israelmcmc:
Thanks @scipascal. Can you also please update run_tutorials.yml accordingly so I can run and test it more easily?

scipascal:
@israelmcmc Done

israelmcmc:
Thanks @scipascal. I can confirm that it works on my mac. I'm merging this! Great work.

Also, thanks @GallegoSav for the support and @ckarwin for testing it in GPU as well.

We should also add this to the tutorial index, and commit a version of the tutorial with outputs so the documentations shows the plots. But we can do this is a follow up PR.

Issue #503, opened 2026-03-10
Missing EN line in orientation file


scipascal:
The SpacecraftHistory removes the last line from the orientation (.ori) file during loading. Since there is no EN line in `DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori`,  the last time bin is lost, causing some background events to have timetags outside of the time axis. Seems also to cause problems with the SAA cut.

israelmcmc:
While we could fix the file, it seems easier to check in the code if the last line is indeed "EN", and otherwise not remove it.

GallegoSav:
@jdbuhler Seems to be related to `.ori` extension but did you also remove the last value when you converted the ori file into fits file ?

jdbuhler:
I parsed the .ori file to create the SpacecraftHistory and then called write_fits() to write the FITS table. So whatever that code path does is what I did. I think the the .ori reader strips the last value, and the FITS writer puts a dummy value back in?

GallegoSav:
@jdbuhler The fact is now there is no more `EN` at the end of `DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori` so it removes the last time value

jdbuhler:
I commented on this in the PR.  The FITS files I generated from DC3_final_* are clearly wrong because they dropped the last line, so I'll need to regenerate.  But are we sure we want to accept .ori files without a trailing EN marker?  This seems like we're inviting future unexpected behavior.

jdbuhler:
I have uploaded fixed FITS orientation files to the develop tree on wasabi and issued PR #517 with the corresponding notebook checksum changes.  I have not touched the parsing code in SpacecraftHistory.

PR #516, opened 2026-03-13
Modify ori file reader to deal with absence of EN (or not) at the end of the file

GallegoSav:
This pr is for solving #503 .

If there is EN at the end of the ori file, `dropna` will remove it. If not , nothing will change

### DIFF
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -410,7 +410,9 @@ class SpacecraftHistory:
         df = pd.read_csv(file, sep=r"\s+", skiprows=1,
                          usecols=tuple(range(1,10)),
                          header = None, comment = '#')
-        vals = df.values[:-1].transpose()
+        #if there is EN at the end of the ori file, then the last line will
+        #contain NaN values.
+        vals = df.dropna()
 
         # assign units to read values
         time_stamps = Time(vals[0], format="unix", copy=False)
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -412,7 +412,7 @@ class SpacecraftHistory:
                          header = None, comment = '#')
         #if there is EN at the end of the ori file, then the last line will
         #contain NaN values.
-        vals = df.dropna()
+        vals = df.dropna().transpose()
 
         # assign units to read values
         time_stamps = Time(vals[0], format="unix", copy=False)
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -412,7 +412,7 @@ class SpacecraftHistory:
                          header = None, comment = '#')
         #if there is EN at the end of the ori file, then the last line will
         #contain NaN values.
-        vals = df.dropna().transpose()
+        vals = df.dropna().values.transpose()
 
         # assign units to read values
         time_stamps = Time(vals[0], format="unix", copy=False)


jdbuhler:
If the issue is that one particular ori file is malformed because the expected EN got dropped, why are we not fixing the file in question, instead of complicating the parser in a way that invites future bugs?

jdbuhler:
To clarify, should we not throw an exception if the EN line is missing?

jdbuhler:
Further comments:

Of the three .ori files used in the tutorials,

* 20280301_3_month... has a trailing EN line
* DC3_final_..._15sbins_... has no trailing EN (but does have a blank line at the end, which the CSV parser ignores) and so was having the last line dropped
* DC3_final_..._1sbins_... has neither a trailing EN nor a blank and so was having the last line dropped

The test case .ori file I recently extracted from DC3_final_..._1sbins_... follows its format and so also lacks any trailing marker. The other three test case .ori files all have an EN marker.

So before trying to fix the code, can we clarify what we consider the correct .ori format?  Can MEGAlib itself ever produce an .ori file without the EN marker?  If not, I would argue that we should fix the two files we are using without such markers and then have cosipy throw an error if the EN is missing.  We could even expend a bit of extra time to check for the EN explicitly, rather than relying on an indirect test with the CSV parser, since .ori is no longer our primary input format now that we have FITS.

In any case, I clearly need to regenerate the FITS file for _1sbins and _15sbins so they have the last line, and also to fix that one test case .ori file.

jdbuhler:
Another comment on how we parse .ori files: the basic parsing call is

df = pd.read_csv(file, sep=r"\s+", skiprows=1, usecols=tuple(range(1,10)), header = None, comment = '#')

Do we actually want to support commenting out lines with '#'?  Is this a feature of MEGAlib's ori files?  If not, I would argue that it is more likely to cause unexpected behavior than to be useful.

israelmcmc:
> So before trying to fix the code, can we clarify what we consider the correct .ori format? Can MEGAlib itself ever produce an .ori file without the EN marker? If not, I would argue that we should fix the two files we are using without such markers and then have cosipy throw an error if the EN is missing. We could even expend a bit of extra time to check for the EN explicitly, rather than relying on an indirect test with the CSV parser, since .ori is no longer our primary input format now that we have FITS.

I know I was the one who suggested the workaround that @GallegoSav implemented in this PR (thanks @GallegoSav btw). I was trying to save some time, but @jdbuhler, if you are willing to fix the files (it seem you already did #517, correct?) and add a check for `EN`, then I agree and we can go with your solution instead. 

> Do we actually want to support commenting out lines with '#'? Is this a feature of MEGAlib's ori files?

I don't think MEGALib comments out lines with `#` for .ori files specifically (it does for other types of files) but we sometimes add information on the files with `#` after the fact manually (for e.g. caveats, provenance, etc.)

jdbuhler:
Yes, I fixed the two offending tutorial .ori files (which are now in the develop tree on wasabi).  Are we good with replacing the existing files in DC3 with these?

I can work on the explicit EN check outside the framework of CSV parsing -- I'll just seek to the end of the file and look there.

israelmcmc:
> Are we good with replacing the existing files in DC3 with these?

My preference is to leave those as is, and only fix the develop and DC4 folders. For DC3 I try to do hot fixes only when absolutely necessary. In this case the error from missing the last line is pretty small, and it doesn't prevent the code from running. Otherwise we would have to change the checksum of the release as well, not just develop. I'm also tagging @ckarwin to see what he thinks.

> I can work on the explicit EN check outside the framework of CSV parsing -- I'll just seek to the end of the file and look there.

Sounds good, thank you.

jdbuhler:
@israelmcmc and @GallegoSav , please see PR #533.

israelmcmc:
I'm closing this PR in favor of #533 + fixing the files in wasabi by @jdbuhler.

Thanks anyways @GallegoSav!

PR #517, opened 2026-03-15
Use fixed FITS reponse files to address issue #503

jdbuhler:
This patch is associated with correcting the FITS orientation files in response to issue #503.  The FITS DC3 orientation files used by the tutorials were missing their last line due to a bug in SpacecraftHistory's .ori reader.  I have updated these FITS files on wasabi to include the missing last line and also uploaded new, fixed versions of the offending .ori files (with EN at the end) to the develop tree; I recommend that we replace the DC3 .ori files with the latter.

The patch itself updates the tutorial checksums for the new FITS files and also corrects a missing EN at the end of one of the test case .ori files.

This patch does nothing to try to fix the parsing of .ori files, as that is being addressed in a separate PR.

After this patch, both the test cases and notebooks all pass.


israelmcmc:
Thanks @jdbuhler. I'm merging and we can continue the discussion about the parsing in #503 and #516

PR #518, opened 2026-03-15
* Store polarization axes in response as type PolarizationAxis

jdbuhler:
This PR changes the FullDetectorResponse to store an explicit polarization angle (PA) convention when it is needed.  That includes both for responses with a Pol axis and for responses using relative CDS coordinates, where the PA convention is needed to convert absolute to relative CDS coordinates. The principal motivation of this change is to avoid having to specify a polarization convention every time we open the response.  (You can still specify the convention, but only as a backup when opening legacy response files that need but don't have a convention. If the convention is specified in the response file, any passed-in convention is ignored.)

The convention of a response is now available via the .pa_convention property, which will be None if no convention is or needs to be defined and is otherwise a PolarizationConvention object.  For responses with a Pol axis, that axis is now presented as a PolarizationAxis, rather than a generic Axis, with a convention that matches the .pa_convention property. (This is true even for HDF5 polarization responses that were written with Pol as a generic Axis; such responses require passing in a convention at opening time.)

In MEGAlib .rsp files, we expect the pa_convention to be specified when needed with a header "PO <convention>".  It is still possible to pass in a convention explicitly if it is needed but not present in the .rsp file; again, any convention present in the file overrides the passed-in convention.  In a tangentially related change, .rsp to HDF5 conversion now parses each MEGAlib axis's AN line to obtain the unit of the axis, rather than assuming fixed units as it did before.

The test_polarization_response files in test_data have been converted to the new format, and the polarization fitting test cases have been changed to stop passing pa_convention, assuming that convention is now available from the response.

Storing a PolarizationConvention object in the FullDetectorResponse tickled some weird behavior in PolarizationAngle / PolarizationConvention.  In particular, PolarizationAngle.transform_to(conv, *args) behaves differently depending on whether conv is a string (a new convention will be created with *args) or a PolarizationConvention object (will be returned unchanged, ignoring *args).  After some discussion with @israelmcmc , I changed the API of these two classes to behave more consistently.  These changes should have zero impact outside of the PolarizationAngle/PolarizationAxis classes, other than the one place I fixed in FullDetectorResponse.get_point_source_response() (which broke under the old API).

NB: the worst of the coverage failures is due to a lack of test cases for the polarization Stokes analysis.  Hopefully I'll be able to address that in the future when I have time to complete a performance overhaul of that analysis.


israelmcmc:
Thanks @jdbuhler 

> If a response is loaded in HDF5 format and has a Pol axis that is not in PolarizationAxis (the old, legacy format), we convert it to a PolarizationAxis using the convention passed to FDR.open() (or the default "RelativeX" if no convention was specified). 

If the input file has the old format, and the user does not specify the convention, can we raise an exception instead? I think that's the current behavior, but even if it's not, it looks safer.

jdbuhler:
I changed the loading behavior for the FDR to behave as you specify.

israelmcmc:
Thank you!

jdbuhler:
RspConverter has been updated to Andreas' specification for the PO header in the .rsp file.

jdbuhler:
@eneights  , I had kicked this PR to you to fix up the polarization analyses, but I've now had time to work through the reasons that they were failing after my changes to store the pa_convention in the FullDetectorResponse.  I think we are good now, but please take a look.

eneights:
This looks good to me and the tutorial notebooks run successfully. Is there now a response file on wasabi containing the polarization convention in the header? If so I think we should update the tutorials to use this and remove the user-defined polarization convention.

jdbuhler:
I don't know that such a response was uploaded, since the patch had not gone in.  Now that DC4 is out, we can restart the develop folder on wasabi and put new responses there.  I can work on that early next week.

eneights:
Ok sounds good! I think this is ready to be merged, but I don't think have permissions to approve/merge it

jdbuhler:
@eneights , just FYI, uploading a modified response is blocked at the moment because the develop data file tree on wasabi was referenced in the tutorials and dos for the 0.4.x releases of cosipy.  Hence, I have to be careful about accidentally overwriting a response file that is in that tree because it might break the release.

I am working on moving files around and removing references to the develop tree from the tutorials and docs, so that (as originally intended) we can freely overwrite files in that tree as part of our ongoing development work.  Hopefully that work will get done in a day or two, but we then have to figure out with @israelmcmc when it will be pushed out as part of a release so we can begin modifying the develop tree again.

israelmcmc:
Thanks @jdbuhler and @eneights  for working on this. 

> but we then have to figure out with @israelmcmc when it will be pushed out as part of a release so we can begin modifying the develop tree again.

I know I already replied on Slack, but just for the GitHub record, any change that only touches documentation can be deployed without a full release. I can push it directly to the branch that ReadTheDocs uses.

@eneights 
> I don't think have permissions to approve/merge it

You should be able to. In the "Files changes" tab you can submit a review approving this PR, which will then allow you to merge it. You can go ahead, and I can take a look at a later time and work on a new issue/PR if needed.

eneights (approved):
Looks good!

PR #519, opened 2026-03-16
Interfaces continuum bg

ckarwin:
Reopening PR.


israelmcmc:
I tracked down the issue with [unit_tests_ml_macos](https://github.com/cositools/cosipy/actions/runs/23165721034/job/67305034664?pr=519) to the fact that both pytorch and healpy are shipped with their own compiles OpenMP libraries, which end up colliding. See https://github.com/pytorch/pytorch/issues/44282. 

The unit test fail in the imaging deconvolution module during a call to `hp.smoothing`. This is a minimal reproducible example:

```
import torch
import healpy as hp
import numpy as np

m = np.ones(hp.nside2npix(32))
hp.smoothing(m)
```

I'm importing pytorch to reproduce the problem, but as you can see it's not really needed for this example. If I install both pytorch and healpy from conda, things work just find. If I install them with pip, I get:

```
OMP: Error #15: Initializing libomp.dylib, but found libomp.dylib already initialized.
OMP: Hint This means that multiple copies of the OpenMP runtime have been linked into the program. That is dangerous, since it can degrade performance or cause incorrect results. The best thing to do is to ensure that only a single OpenMP runtime is linked into the process, e.g. by avoiding static linking of the OpenMP runtime in any library. As an unsafe, unsupported, undocumented workaround you can set the environment variable KMP_DUPLICATE_LIB_OK=TRUE to allow the program to continue to execute, but that may cause crashes or silently produce incorrect results. For more information, please see http://openmp.llvm.org/
```

I then tried to set `KMP_DUPLICATE_LIB_OK=TRUE`, but I'm getting `Segmentation fault: 11`.

I’ll look into the links in the pytorch issue to see if someone has a good workaround. For now, it seems that the only think that consistently works is to install both healpy and torch from conda. Which is not ideal, since it seems to me that pytorch is moving toward supporting mainly the pip installation (they don’t even mention conda in the [installation instructions](https://pytorch.org/get-started/locally/), and the [official anaconda channel was deprecated](https://github.com/pytorch/pytorch/issues/138506))

israelmcmc:
@ckarwin As I mentioned during the call, I couldn't find a nice/robust way to catch if the user had a duplicated OpenMP libraries. However I implemented an, I think, OK solution. Please see my PR to your branch [here](https://github.com/ckarwin/cosipy/pull/7): 

1. I moved `ContinuumEstimationNN` to `cosipy.background_estimation.ml`. The `.ml` submodule at the end serves two purposes 1) pytorch is not imported by default with the rest of the classes, which prevents it from clashing with other module (unless the user explicitly imports it). This fixed the image deconvolution unit tests.  2) `.ml.__initi__.py` checks if pytorch is installed, so all the classes inside it don't need to do the check themselves.

2. I made `ContinuumEstimationInterp` the base class, and `ContinuumEstimationNN` a derivative. That way `ContinuumEstimationInterp` can live outside `.ml` and be used even without torch.

3. I added a flag to know if cosipy was installed with `[ml]`. You can do `import cosipy`/`cosipy.with_ml`, which is True to False.

4. I added `[ml]` to the installations instructions, as well as the OpenMP workaround to the troubleshooting instructions. 

5. I updated the tutorial and fixed an absolute path so it can run automatically.

6. Sync your branch with develop and fixed from conflicts.


With these change both the tutorial and the unit tests work for me. 

Please take a look at my PR and if you are OK, please merge it and we can finally get this PR merged --what a tongue twister!

ckarwin:
Thanks, @israelmcmc. I think your changes look good. Yes, it turned out to be more complicated than we had anticipated!  

I ran into a couple of issues when I was reviewing your PR. 

I installed your branch and tried to run the tutorial from the automatic pipeline. It failed, and it seems like the issue is that the ori file doesn't have "EN" at the end. I believe this is coming from the second part where the spectral fit is performed. I couldn't see exactly what the error was from the log output, and so I will look into it today. I know there were some changes recently relating to the ori file. Did you encounter this problem?

I think making `ContinuumEstimationInterp` the base class is fine. However, I see that it still contains the `plot_training_loss `method. This should probably be moved to the NN class, since it's only used with the NN.

israelmcmc:
@ckarwin Thanks for checking it.

For the EN thingy, you probably just need to redownload the file. It was recently updated, see https://github.com/cositools/cosipy/pull/533. I run it with run_tutorials.py, which makes sure the checksum is the same.

```
~/software/cosipy/docs/tutorials/run_tutorials.py ~/software/cosipy/docs/tutorials/run_tutorials.yml -o tutorials --wasabi_mirror ~/cosi/data/wasabi/cosi-pipelines-public/ --tutorial  background_continuum
```

It makes sense to move `plot_training_loss`, I'll work on it. I missed it because I only looked for methods that used torch,.

After my changes the unit tests are failing (preview on #539), despite they running fine in my mac. I also need to figure this one out.

ckarwin:
@israelmcmc I also run it with `run_tutorials.py`. In `run_tutorials.yml` I see that the ori file is being copied from `COSI-SMEX/DC3/Data/Orientation/DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.ori`. I just checked this file on wasabi and it doesn't have the EN at the end. Is this the same file you're using?

israelmcmc:
@ckarwin I tested the tutorials before merging develop, but pushed the version after the conflict resolution. I think that’s way I was note getting the issue, sorry. I'm fixed the paths for the new ori files Jeremy uploaded (in the develop folder), and I'm running it again. I'll push the changes once I verify it's working.

I think the unit tests are failing because they share the same environment, so torch is still being loaded

This works: `pytest tests/background_estimation/test_line_background_estimation.py`
This works: `pytest tests/background_estimation/test_continuum_background_estimationNN.py`
Both together fail: `pytest tests/background_estimation/test_line_background_estimation.py tests/background_estimation/test_continuum_background_estimationNN.py`

Ugh... I'll think about what to do.

israelmcmc:
@ckarwin I pushed the fix for the tutorial, it's working for me. Can you please give it another try? 

For the record, I changed the .ori file to:
```
# DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.fits
fetch_wasabi_file('COSI-SMEX/develop/Data/Orientation/DC3_final_530km_3_month_with_slew_15sbins_GalacticEarth_SAA.fits', checksum = 'ca94ff1d7a73c1f41479aaf598807673')
```

(new format from jdbuhler)

ckarwin:
Thanks for checking. I updated with your latest changes and am running the tutorial now.

ckarwin:
@israelmcmc Ok, the tutorial succeeds now with your latest changes. I think the only issue now is the unit test.

israelmcmc:
Great! The failure with the LineBackgroundEstimation is also related healpy + torch (it also called hp.smoothing), but I'm trying to figure out why now it no longer affect the imaging test but it does the line bkg ones. I'll check what being imported more carefully.

israelmcmc:
@ckarwin I think I'm done now.

First, this previous statement was wrong:
> I'm trying to figure out why now it no longer affect the imaging test but it does the line bkg ones. 

It was still affecting the imaging test, I was just calling `image_deconvolution/test_image_deconvolution.py` instead of `image_deconvolution/test_model.py`, which is the one that actually uses hp.smoothing. 

So, all tests were passing individually, but they failed when all of them were called together because, again, both healpy and pytorch are imported. I don't think there's a way to avoid this, since pytest imports everything during the initial collection step.

As a workaround, I installed healpy and torch from conda in the `unit_test_ml_macos.yml` action, and documented all this in the installation instructions. The unit tests are now passing #539. Everything seems fine.

Since this was the only remanning issue, and I think you were already OK with all other changes, I'll go ahead and merge this. 

Thanks for this work and the patience, @ckarwin!

ckarwin:
No problem! Thanks for all your work on this! cosipy is growing to be a complex library. Do I still need to merge your PR into my branch?

israelmcmc:
> Do I still need to merge your PR into my branch?

Nope, all the commits are already in develop. Your and my feature branches can be safely deleted.

PR #521, opened 2026-03-18
Pointing aware analysis update to good_time_interval.py

### LINKED ISSUE #391: Pointing-aware point source analysis (binned modeled analysis)
Don't include times when a point source is behind the Earth.

For a binned analysis, this involves a time selection cut.

If the binned analysis does not have time dependence --i.e. it has no time binning-- then this only works for a single source. Having many time bins is likely to make things too slow. See #392.

parshadkp:
Added functionality for pointing aware analysis in good_time_intervals.py. This will get time intervals that are within the source exposure window, taking in account earth occultation if turned on. Also added testing code in test_GTI.

Note that eventually we should move away from max_offaxis (half FOV) to a FOV that is function of energy, which is more realistic.

### DIFF
--- a/cosipy/event_selection/good_time_interval.py
+++ b/cosipy/event_selection/good_time_interval.py
@@ -2,6 +2,8 @@ import logging
 logger = logging.getLogger(__name__)
 
 import numpy as np
+import astropy.units as u
+from astropy.coordinates import SkyCoord
 from astropy.time import Time
 from astropy.io import fits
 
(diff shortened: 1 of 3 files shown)


cosipy/event_selection/good_time_interval.py
```diff
@@ -171,6 +173,64 @@ def from_fits(cls, filename):
...
+        """
+        Build a GTI where a fixed sky position is within a maximum off-axis angle of the spacecraft boresight.
+
+        Parameters
+        ----------
+        target_coord : astropy.coordinates.SkyCoord
+            Fixed target (source) position.
+        sc_history : cosipy.spacecraftfile.SpacecraftHistory
+            Spacecraft pointinghistory to evaluate (.ori file).
+        max_offaxis : astropy.units.Quantity or float
```
israelmcmc: Can this be only Quantity objects, please?

I think accepting bare values and default units is error-prone. In some methods in cosipy that require high performance I've put the unit in the parameter name e.g. `max_offaxis_deg`, but I don't think this particular function will benefit much from this.

israelmcmc:
Thanks @parshadkp! I still need to test it, but the code looks good. So far just, the detail about the units I commented on.

> Note that eventually we should move away from max_offaxis (half FOV) to a FOV that is function of energy, which is more realistic.

I think this is harder than it sounds. The reason is that you need to select events best on the _measured_ energy (Em). You can't cut on the _real_ energy (Ei), since you don't know it, and you would still need to integrate over Ei while convolving your response. Cutting in Em effectively changes the normalization and distribution of the response, so you would have to either create a response for a particular source, or make this correction on the flights.

Note that in the unbinned analysis (or equivalently, if you were to have very fine time and Em bins) your sensitivity won't benefit from a cut on the field of view and earth occultation, even if they are a function of Em. This is because each narrow time/Em bin is containing it's own signal and background. For example, a bin with 0 expected signal has the same effect as removing it. In fact, you gain sensitivity by not applying the cut at all --i.e. the unbinned analysis effectively does a sort of "event weighting", as opposed to a sharp on/off cut. 

To clarify, this cut is useful:
- When integrating all of your data into a single time bin (or a few bins). Applying this cut indeed effectively removes background
- For the unbinned case, even if you don't benefit in sensitivity, it's useful to apply this to avoid spending computing resources for events that don't add much sensitivity.

parshadkp (commented):
max_offaxis now changed only takes astropy.units.Quantity.

cosipy/event_selection/good_time_interval.py
parshadkp: Fixed in latest commit.

israelmcmc:
Thanks @parshadkp. Question: Are the private function `sc_history._` really necessary here? Is it significantly slower without them? If so, I'll check tomorrow if I can work on the SpacecraftHistory to fix this. The functions with the underscore are mean to be used only by the SpacecraftHistory class itself.

parshadkp:
Hi @israelmcmc. I could get rid of sc_history._get_target_in_sc_frame but that would require adding a few more lines to code to calculate the angle between the pointing and the source. This is not too hard to do but since sc_history already does this, I thought it would be easier this way. Let me know if you'd like me to change this. 

I do not know any other way to see whether the source is occulted without sc_history._get_earth_occ. If you know of one, I'm happy to change it to that.

israelmcmc:
@parshadkp I ended up creating public versions of the function you need, moving some lines in your code to the new methods. I modified how you created `DummySpacecraftHistory`, but the values are the same and the test are still passing. I'm merging! Thanks for adding this.

PR #523, opened 2026-03-19
Fix tests/dataIO/test_unbinned_data_all.py in Fornax

israelmcmc:
Change scp and gunzip call using os to shutil and gzip library combo.
- Motivation: test was failing in a system where scp was blocked. The copy is local, so the 's' was not needed


israelmcmc:
I'm merging this as a maintenance task.

@ckarwin just FYI

PR #525, opened 2026-03-19
Add MDP and Stokes parameters fitting for maximum likelihood method

eneights:
I added a function to calculate the minimum detectable polarization for the maximum likelihood method, and added the calculation to the example notebook. This runs quite slowly, and should probably be parallelized. The notebook only runs 100 iterations for now, but in reality we will want something like ~10,000. It also currently can only handle a single source and background component.


cosipy/sensitivity/mdp.py
```diff
@@ -0,0 +1,129 @@
...
+	likelihood method.
+
+	Parameters:
+	n : int
+		Number of simulations.
+	source_direction : :py:class:`astropy.coordinates.SkyCoord`
+		Source direction.
+	spectrum : :py:class:`threeML.Model`
+		Source spectrum.
+	model : :py:class:`astromodels.core.model.Model`
```
israelmcmc: It seems to me that you can change `(spectrum, model, source_direction)` for a single input call e.g. `source` that received a `PointSource`. You can get all of the information above from there and at the same time makes sure you only have a single source.

israelmcmc:
Thanks @eneights. 

I added the `__init__.py` file to the new `sensitivity` folder is recognized as a module. It needs one even if it's empty. I also synced your branch with develop.

Issues/requests/comments:

* Since now `sensivitiy` is recognized as a module, codecov is correctly catching that you haven't written unit test for it yet. 

* The tutorial is complaining the `compute_mdp` was not imported. For reference, you can check that the tutorial works in a clean run with

```
~/software/cosipy/docs/tutorials/run_tutorials.py ~/software/cosipy/docs/tutorials/run_tutorials.yml -o tutorials --wasabi_mirror ~/cosi/data/wasabi/cosi-pipelines-public/ --tutorial polarization_ml
```

* Have you looked into the fail fits? 20% failure is like a large fraction, considering that the MDP is based on the ~99%, if those simulations that would have had a large fake PD are more likely to fail, then the MDP could be  underestimated.

* It seems to me that you can change `(spectrum, model, source_direction)` for a single input e.g. source that received a `PointSource`. You can get all of the information from these 3 parameters from there and at the same time makes sure you only have a single source.

* Similarly, you don't need `fix_bkg`, since you can pass this information through `bkg_parameter` e.g.
```
Parameter('total_bkg',  # background parameter
                                            0.0016,  # initial value of parameter
                                            min_value=0,  # minimum value of parameter
                                            max_value=100,  # maximum value of parameter
                                            delta=0.05,  # initial step used by fitting engine
                                            unit = u.Hz,
                                            free = False)
```

A note for later: I think that eventually this function need to be split in two: a generic one that takes `(source: PointSource, background:BackgroundInterface, response:ThreeMLSourceResponseInterface)`, and a helper function that builds `(source, background,response)` from the specific implementation you currently have. However, this PR made me realize that I first need to change slightly the definitions of some interfaces to make this possible.

eneights:
Thanks, @israelmcmc! I'm not sure what to make of the fits failing and it does make me worry the MDP is being underestimated. But I'm not sure what to look into to see why they're failing, since they're all unpolarized and the only difference between them is the Poisson fluctuations. 

About changing (spectrum, model, source_direction) for a single input, I had tried quickly before to only have the model as an input since this contains the spectrum and source_direction, but I couldn't figure out how to get the spectrum object out instead of just the dictionary of parameters. I'll look into it more a little later.

eneights:
It's updated now so the user only has to input the model and not the spectrum and position in addition

cosipy/response/instrument_response.py
```diff
@@ -294,9 +289,36 @@ def rotate_coords(c, rot):
...
+
+            # wrap 180-degree polarization angles to keep them
+            # within bin range
+            la = loc_pol_angles.angle
+            la = np.where(la.deg == 180., 0. * u.deg, la)
+
+            # map each local-convention Pol bin angle to
+            # nearest bin (TODO: this could also be
+            # interpolated)
+            loc_pol_bins = self._dr._axes['Pol'].find_bin(la)
```
israelmcmc, line 311: I'm thinking that using a 0-th order approximation might be throwing the minimizer off --i.e. the likelihood vs PA is constant within a bin, and then suddenly jumps. Since the minimizer typically depend on, at least, the first derivative, this sort of jumps cause convergence issues most of the time.

It might be worth doing at least a first interpolation, it's not that much harder. Histpy already supports getting these weights.

cosipy/response/PointSourceResponse.py
```diff
@@ -79,14 +81,30 @@ def get_expectation(self, spectrum, polarization=None, flux=None):
...
+
+        if isinstance(polarization, StokesPolarization):
+
+            Q = polarization.Q.value.k.value
+            U = polarization.U.value.k.value
+            pd = np.sqrt(Q**2 + U**2)
+
+            if pd > 1.:
+
+                factor = np.exp((pd - 1.)**2)
```
israelmcmc, line 96: I understand this fudge factor is to bring back the minimizer back to PD<1 by artificially making the likelihood really bad beyond that. This seems to be failing though, but not by itself though. When the likelihood is flat vs a given parameter (see my comment about interpolating above) the minimizer send the next iteration very far away, since the derivative is 0. It seem that it is landing beyond the PD for which `np.exp((pd - 1.)**2)` is well defined numerically, turning into `inf` and then `nan`, and therefore messing up the convergence.

cosipy/sensitivity/mdp.py
```diff
@@ -0,0 +1,135 @@
...
+	failed_fits = 0
+
+	if response_pa_convention:
+		dr = FullDetectorResponse.open(response_file, pa_convention=response_pa_convention)
+	else:
+		dr = FullDetectorResponse.open(response_file)
+	
+	for i in range(n):
+
+		this_model = copy.deepcopy(model)
```
israelmcmc, line 63: This `this_model` copies `model`, and it is then as the starting point below in `JointLikelihood(this_model, ...)` the initial guess during the fit is always the same as the one by `model`. In principle the solution shouldn't depend on the initial condition, but it would be good to check this, or maybe randomize the initial guess so it doesn't have an effect on the MDP

cosipy/response/PointSourceResponse.py
israelmcmc: Note: the 🤖 also (correctly) pointed out (A3) that since this is fudge factor on the overall expectation, the background can in principle be floated to more or less cancel out this fudge factor. I think it should still work, since the background has a different shape, but it's worth keeping this in mind. If after solving the other issues we still have problem fitting Q/U, I think the real solution will be to come up with a new astromodels polarization re-parametrization that sets PD=1 at infinity.

cosipy/response/instrument_response.py
```diff
@@ -237,7 +233,7 @@ def _differential_effective_area_inertial(self,
                                pol_edges,
                                pol_edges[-1] + 2 * (center_angles[-1] - pol_edges[-1]) ]
 
-            out_axes.append( PolarizationAxis(pol_edges, unit=u.deg, convention = polarization.convention, copy=False) )
+            out_axes.append(PolarizationAxis(pol_edges, unit=u.deg, convention='iau', copy=False, label='Pol'))
```
israelmcmc, line 236: You hardcoded the convention to iau, but this is called through
`PointSourceResponse.from_scatt_map(..., polarization_axis, ...)`->`BinnedInstrumentResponse.differential_effective_area(..., polarization...)`->``BinnedInstrumentResponse._differential_effective_area()`. The last one drops the polarization parameter, which can be different than IAU.

cosipy/sensitivity/mdp.py
```diff
@@ -0,0 +1,135 @@
...
+		Path to response file.
+	response_pa_convention : str, optional
+		Polarization angle convention of response.
+	response_frame : str, optional
+		Frame of response file.
+	confidence : int, optional
+		Confidence level.
+	"""
+
+	if len(model.source.spectrum.to_dict()) > 1:
```
israelmcmc, line 45: `source` is not a attribute of `astromodels.core.model.Model`, just an arbitrary source name. In this case I'd take a `Source` as argument directly instead of a `Model`.

cosipy/sensitivity/mdp.py
```diff
@@ -0,0 +1,135 @@
...
+
+		plugins = DataList(cosi)
+
+		like = JointLikelihood(this_model, plugins, verbose=False)
+
+		try:
+
+			_ = like.fit(quiet=True)
+
+		except:
```
israelmcmc, line 118: It's better to except a particular error (the one for which you wrote the except for) than a blanket except, which can hide other problems

cosipy/sensitivity/mdp.py
```diff
@@ -0,0 +1,135 @@
...
+
+	degrees = []
+	failed_fits = 0
+
+	if response_pa_convention:
+		dr = FullDetectorResponse.open(response_file, pa_convention=response_pa_convention)
+	else:
+		dr = FullDetectorResponse.open(response_file)
+	
+	for i in range(n):
```
israelmcmc, line 61: Most objects in this loop are the same for all iterations. They can be taken out of the loop, and inside the loop only for Poisson fluctuations + fit. This should make things significantly faster.

israelmcmc (changes requested):
Thanks @eneights! I'm sorry about the long delay!

I instructed Claude Code to review this PR. I then extracted the bits that I think matter the most (particularly the ones that I think are messing up the fit in addition to the coarseness of the response) and "translated" them into the comment that I posted here.

I won't pushed at this time for all changes suggested by Claude Code, but you can see them here:
https://github.com/israelmcmc-ai/cosipy/blob/ce2c5fffa5b5e267743a3674c3b63d089d0f5cab/PR525_REVIEW.md

It think they make sense. Many of these are not issues now, but might come bite us later.

PR #532, opened 2026-03-20
Proactively add **kwargs to Histogram subclasses' _open() and ._write() methods

jdbuhler:
This patch changes the two response classes that subclass Histogram and override its _open and/or _write methods to accept **kwargs and pass them on to the superclass.  This change is harmless right now and will support planned extension of the base Histogram class's _open/_write to support compression.


israelmcmc:
Thanks @jdbuhler. I agree it should be benign, so I'm merging.

For reference, see https://gitlab.com/burstcube/histpy/-/merge_requests/16

PR #533, opened 2026-03-21
Explicitly check for "EN" at end of .ori files

### LINKED ISSUE #503: Missing EN line in orientation file

### LINKED ISSUE #527: Fix lonlat bug in class SpacecraftHistory.get_exposure
lonlat should be False here.
https://github.com/cositools/cosipy/blob/3f119161c2e797ad624d13ba5b3e21a62b41fb81/cosipy/spacecraftfile/spacecraft_file.py#L1098

Bonus point if you add a regression unit test.

jdbuhler:
MEGAlib .ori files end with the tag "EN" on a line by itself.  This tag is removed when the file is read in.  Recently, we discovered that some .ori files in cosipy were missing that terminal EN; those files were fixed, but we still need to properly detect the missing EN.

This patch chooses to reject files without the terminal EN, rather than attempt to work around them, because trying to "do the right thing" would need to handle all sorts of cases.  For example, there might be multiple lines of garbage at the end of the file.

This PR is intended to supersede #516.

Since I was editing this file anyway, I also included a one-line fix for issue #527 with a suitable test case.

### DIFF
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -402,10 +402,25 @@ class SpacecraftHistory:
         # Type OrientationsGalactic
         # ...
         # EN
-        # Using [:-1] instead of skipfooter=1 because otherwise it's
-        # slow and you get ParserWarning: Falling back to the 'python'
-        # engine because the 'c' engine does not support skipfooter;
-        # you can avoid this warning by specifying engine='python'.
+
+        # Check that the EN line is present (and throw an error if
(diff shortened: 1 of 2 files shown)


israelmcmc:
Thanks @jdbuhler. Looks good, I'm merging.

PR #536, opened 2026-03-25
Add empty pipeline/__init__.py 

israelmcmc:
So pipeline is detected as a module and included in the coverage


israelmcmc:
Wrong location!

PR #543, opened 2026-04-02
Faster TimeSelector and SpacecraftHistory

israelmcmc:
TimeSelector:
- Faster, using only 64bit precision
- Add unit tests
- Be consistent about tstart inclusive, stop exclusive.

I'll work on a similar trick (64bit precision) for SpacecraftHistory (I'll abandon histpy.TimeAxis in favor of a regular Axis since COSI doesn't need <1us precision).

### DIFF
--- a/cosipy/event_selection/time_selection.py
+++ b/cosipy/event_selection/time_selection.py
@@ -9,12 +9,12 @@ from astropy.time import Time
 
 from cosipy.interfaces import TimeTagEventInterface, EventInterface, TimeTagEventDataInterface
 from cosipy.interfaces.event_selection import EventSelectorInterface
-from cosipy.util.iterables import itertools_batched
+from cosipy.util.iterables import itertools_batched, asarray
 
 
 class TimeSelector(EventSelectorInterface):
 
-    def __init__(self, tstart:Time = None, tstop:Time = None, batch_size:int = 10000):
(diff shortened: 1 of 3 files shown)


jdbuhler (commented):
I don't have a good sense of how expensive the conversions between Time and Julian date are, but I assume they are way less painful than the TimeAxis operations.  I'll have to benchmark the SpacecraftHistory after the changes to see what the overhead of select_interval() is.

I see that intervals_duration() now returns a value in days.  Are there other places where we were used to using seconds as the typical time unit of orientation history but are now going to have to change?

israelmcmc:
@jdbuhler Thanks for checking. I used Julian days because that's what the underlying Time objects use. It just needs to sum jd1+jd2. Hopefully this sum is not much of an overhead, and it will allows us to still use Time for the inputs/outputs.

> I see that intervals_duration() now returns a value in days. Are there other places where we were used to using seconds as the typical time unit of orientation history but are now going to have to change?

No code should assume units of intervals_duration(). The function only "promises" to return Quantity. If we find such instance using `.value`, we should change them to `to_value(u.s)` (if really needed in seconds). I also used days here to avoid the extra conversion from the underlaying jd

jdbuhler (approved):
OK, approving review

israelmcmc:
Thanks @jdbuhler !

Issue #550, opened 2026-04-07
Selecting a sub-interval of a SpacecraftHistory does not preserve the original’s cache_earth_occ settings


israelmcmc:
Noted by @jdbuhler

jdbuhler:
Addressed by PR #551

PR #551, opened 2026-04-07
Fix issues related to earth occultation in SpacecraftHistory

jdbuhler:
This patch addresses some issues that arose when exercising the earth occultation code in SpacecraftHistory:

(1) A test-case ori file, 20280301_2s, assigned zero altitude to each orientation.  After conversion to GCRS, this caused SpacecraftFile to compute earth-centered distances less than the fixed earth radius used for occultation testing, which leads to nonsense (NaN) results.

(2) In the hopefully unlikely case that we ever have the satellite orbiting at near-zero altitude, we should (a) duck and (b) clamp the computed earth-centered distance to be at least the earth radius, with a warning.

(3) SpacecraftHistory methods that subset the orientation list, such as select_interval(), do not preserve earth_occ caching behavior relative to the parent orientation.  This is issue #550 .

This patch fixes the test case in (1), adds the sanity check in (2), and transfers the caching behavior in (3) for both select_interval and apply_gti.  I've added test cases for the caching behavior and for an explicit, nontrivial occultation calculation.


israelmcmc:
Thanks @jdbuhler. This looks good to me. I added a test for apply_gti in #552. Please take a look and if it looks good to you please approve it and merge. Then you can add the extra test that you mention to this PR and I can merge it.

jdbuhler:
I merged the gti test case and augmented it to check that caching behavior is preserved.

israelmcmc:
Thanks @jdbuhler . Merging!

PR #554, opened 2026-04-09
Remove test case detritus from continuum background estimation

jdbuhler:
Continuum background estimation writes a Histogram, inpainted_estimated_bg.h5, whenever it runs.  The default behavior is to leave this file in the current working directory.  For testing, it should be directed to the temporary directory so as not to pollute the CWD.

This patch redirects where the write happens in the two relevant test cases.


israelmcmc:
Good catch, I agree. Thanks @jdbuhler

Issue #358, opened 2025-05-23
Allow BinnedData to track under/overflow


israelmcmc:
Histpy version 1.x had the under/overflow bins turned on by default, so so no event was ever discarded. Version 2.x has these turned off by default (for optimization purposed) and throws a warning when some are discarded. 

Sometimes this is unexpected, and an user doing an analysis would benefit by inspecting the under/overflow bin to check is they are as (presumably small) as expected. 

This line would need to change:
https://github.com/cositools/cosipy/blob/0570719bcc7ed1735ce6d5eadbae3b45b24f99ac/cosipy/data_io/BinnedData.py#L110

jdbuhler:
I suggest adding a boolean flag to get_binned_data() (defaults to False) that would enable overflow tracking in the binned array.

I would also suggest adding a flag to control whether the output histogram is sparse, and consider whether the default should be True or False.  Dense binning is likely significantly faster than sparse binning.

PR #555, opened 2026-04-10
enable overflow/underflow tracking in BinnedData histogram 

ashwin230998:
self.binned_data = Histogram(axes, sparse=True, track_overflow=True)
Added track_overflow parameter


israelmcmc:
Thanks @ashwin230998. Can you please add the flags suggested by @jdbuhler in #358?
https://github.com/cositools/cosipy/issues/358#issuecomment-2905024880

Let's go with defaults  sparse=None and track_overflow=None to match the histpy defaults.

Issue #504, opened 2026-03-10
Arithmetic with Time and TimeDelta Objects is too slow


scipascal:
Performing operations like `np.searchsorted` with `Time` objects is significantly (several orders of magnitude) slower than using the underlying values (e.g., `.utc.unix`). This is the bottleneck in both the `SpacecraftHistory` interpolation, which relies on histpy `self._livetime_hist.axis.interp_weights_edges(times)`, as well as in the `TimeSelector`.

israelmcmc:
Thanks @scipascal.

Tagging @jdbuhler, since he's looked into this. 

My understanding is that so far the best course of action:
- In histpy: implement a derived class from TimeAxis called e.g. TimeAxis64, which internally stores the time in a single double number, as opposed to the 128 bit precision that astropy has.
- In cosipy: as a workaround, bypass histpy's interpolation and do it directly in cosipy with a double.

PR #566, opened 2026-04-14
Performance improvements to SpacecraftHistory

jdbuhler:
Following the recent PR #504, which addressed a serious performance bug with time axes, SpacecraftHistory is now usable for applications that exercise the time axis, in particular select_interval().  However, benchmarking revealed that there are still some performance issues worth fixing.  In my usage of SpacecraftHistory with ts_map code, I found that the amount of time spent calling select_interval() to trim the orientation history for each of the 17 DC3 simulated transients was greater than the total time needed to map them all.  This was not the case with the old SpacecraftFile class.

This patch addresses several issues that were causing unnecessary slowdowns.  The big one is that select_interval() should not perform operations on the entire orientation history when it could do so for just the subset of interest.  A call to attitudes.as_matrix() on the whole Attitude array was the worst offender -- particularly because the conversions are being done just to facilitate concatenation of Attitudes.  The revised code avoids conversions by concatenating the Attitudes' underlying Rotations together directly.  A second issue is that live time is being stored as a 1-D Histogram; extracting the contents from this Histogram every time it is needed added a surprising amount of overhead, which we remove by keeping livetime as a bare Quantity (but still keeping the useful time Axis object).

One issue that was not in cosipy itself is that calling interp_edges() on a large Axis object takes time proportional to the number of edges on the Axis, even if only one point is being interpolated.  This was a histpy problem that I have addressed through a separate MR against that package.

### DIFF
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -803,6 +803,8 @@ class SpacecraftHistory:
 
         """
 
+        from scipy.spatial.transform import Rotation as R
+
         if start is None:
             start = self.tstart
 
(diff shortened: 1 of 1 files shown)


israelmcmc:
Thanks @jdbuhler. 

Just one thing: I'd change self.time_axis to self._time_axis to hide this detail of the implementation (which might change)

jdbuhler:
Done.

israelmcmc:
Thank you! Merging!

PR #568, opened 2026-04-15
Fixes for recent notebook testing failures

jdbuhler:
This patch fixes two failures observed when testing the tutorial notebooks:

(1) SpacecraftHistory fails to load an orientation file if one but not both of tstart and tstop is specified.  Make sure we don't try to use these values in arithmetic if they are None.

(2) The path to the FITS orientation file in ASAD_method.py was updated incorrectly by the recent .notebook fixing PR.


israelmcmc:
Thanks @jdbuhler. Looks good.

I also cherry-picked the changes in `docs` and updated the branch for ReadTheDocs (branch v0.4.2_rdt): https://github.com/cositools/cosipy/commit/fd4b5a4a3d2c271b6153648311b080cff642ecbe

PR #570, opened 2026-04-16
Enable overflow/underflow tracking and dense/sparse binning

ashwin230998:
Enable overflow/underflow tracking and dense/sparse binning. swapped .todense() and contents to enable both dense/sparse binning. (issue #358)

### DIFF
--- a/cosipy/data_io/BinnedData.py
+++ b/cosipy/data_io/BinnedData.py
@@ -30,7 +30,7 @@ class BinnedData(UnBinnedData):
     def get_binned_data(self, unbinned_data=None, output_name=None,
                         make_binning_plots=False, show_plots=False,
                         psichi_binning="galactic", event_range=None,
-                        weights=None):
+                        weights=None, sparse=None, track_overflow=None):
 
         """Bin the data using histpy and mhealpy.
 
@@ -54,6 +54,11 @@ class BinnedData(UnBinnedData):
         weights : value or array of values, optional
             weight to use when filling the Histogram, if not set histpy will
             use weight of 1
+        sparse : bool, optional
+            'True' for sparse binning, or
+            'False' for dense binning (default is False).
+        track_overflow: bool, optional
+            option to track under/overflow bin (default is False).
 
         Returns
         -------
@@ -124,7 +129,7 @@ class BinnedData(UnBinnedData):
                      Axis(energy_bin_edges, unit=u.keV, label='Em'),
                      Axis(phi_bin_edges, unit=u.deg, label='Phi'),
                      psichi_axis], copy_axes=False)
-        self.binned_data = Histogram(axes, sparse=True)
+        self.binned_data = Histogram(axes, sparse=sparse, track_overflow=track_overflow)
 
         # Fill histogram:
         if event_range is None:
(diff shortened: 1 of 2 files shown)


hiyoneda:
@ashwin230998 It seems that you open the same PR several times (#570, #569, #555). Is it possible to close two of them?
Also, could you share the purpose of this change? Actually, enabling overflow/underflow tracking increases memory usage because we need to keep two additional bins for each axis. For this reason, we disabled this tracking before.

ashwin230998:
@hiyoneda sorry for that I closed the previous PR.    This was updated in order to give the user to turn on the overtracking if they want to check it . (Issue #358)

hiyoneda:
I see, thanks. Maybe @jdbuhler or @israelmcmc have comments on this.

israelmcmc:
Yeah, I asked @ashwin230998 to work on #358. As he said, it's just so people can check the under/overflows is they want (a typical analysis check), but the default remains the same. 

The `sparse` parameter default does change, which I also asked @ashwin230998 to modify. At some point we though the sparse this would be helpful, but it doesn't seem to be the case. Most analysis currently convert it to dense anyway. However, the option will stay there if we change our mind.


@ashwin230998: You can go ahead and remove these line to make the ASAD test pass, we no longer need that check.
```
# sparse contents has no unit
assert binned_data.is_sparse
```

jdbuhler:
The default values set for track_overflow and sparse are "None".  For an empty Histogram, that means "follow the histpy default".  I don't expect the defaults (sparse = False, track_overflow=False) to change in future, but do we want to set the defaults to False in the method signature?  (I will add that "None" means something quite different when initializing a non-empty Histogram -- it means "use the properties of the provided data").

Performance observation: calling to_dense() on a Histogram *always* makes a copy, whether or not a sparse-to-dense conversion is necessary.  Hence, changing all the occurrences of ".contents.to_dense()" to ".to_dense().contents" introduces an extra copy in each case.  If we can't use the old ordering (because we don't know for sure that the Histogram is sparse?), it would be worth introducing a trivial wrapper function to avoid copying what might be a very large Histogram.

Possible feature suggestion: as long as we are allowing control of sparse and track_overflow, do we also want to add a 'dtype' argument?  Default None means use the default dtype for histpy (float64); non-default means "use this dtype for the data".  This would be helpful if we want to create data sets with single-precision float for efficiency reasons.

israelmcmc:
> but do we want to set the defaults to False in the method signature? (I will add that "None" means something quite different when initializing a non-empty Histogram -- it means "use the properties of the provided data").

In this case `None` should be fine because it for the Histogram initialization (zeroth out).

> Performance observation: calling to_dense() on a Histogram always makes a copy, whether or not a sparse-to-dense conversion is necessary. Hence, changing all the occurrences of ".contents.to_dense()" to ".to_dense().contents"

Ah good point, I had forgotten about this. I think the cleanest solution would be to add an option `copy=False` to histpy's to_dense and to_sparse (=copy only if needed). That is, performing the check of whether it's already dense in histpy, as opposed to a custom function in cosipy. What do you think @jdbuhler ?

> Possible feature suggestion: as long as we are allowing control of sparse and track_overflow, do we also want to add a 'dtype' argument? Default None means use the default dtype for histpy (float64); non-default means "use this dtype for the data". This would be helpful if we want to create data sets with single-precision float for efficiency reasons.

Yeah, that would be good. Although it doesn't need to be in this PR to merge it.

jdbuhler:
Yes, it would be good to add a copy parameter in histpy though we'll have to leave the default as "True" due safety reasons. I can do that.

israelmcmc:
Thanks @jdbuhler! Yes, True as default makes sense.

jdbuhler:
OK, copy parameter for to_dense/to_sparse is now histpy MR !18.  The conversions themselves are now faster as well.

israelmcmc:
Thanks @jdbuhler. I released histpy 2.0.7 with that change in it, and updated our requirements in #571.

@ashwin230998 Now you can specify `to_dense(copy = False)` to avoid this unnecessary copy, please.

israelmcmc:
Thanks @ashwin230998, and @hiyoneda and @jdbuhler for checking on this. It looks good to me now, I'm merging it.

PR #575, opened 2026-04-16
Utilize new histpy to_dense(copy=False) behavior

jdbuhler:
Histpy 2.0.7 introduced the copy argument to Histogram.to_dense().  Specifying to_dense(copy=False) avoids copying an existing dense Histogram while still doing sparse-to-dense conversions for sparse Histograms.

This patch uses the new to_dense() semantics throughout cosipy, both in the code and in the notebooks.  Almost all uses of to_dense() (or its alias todense()) occur in one of two cases:

* histogram.to_dense().contents -- extracts a guaranteed-dense version of the Histogram's contents. We can safely add copy=False if the original Histogram will not be read after the contents array is written.  This code works whether histogram is dense or sparse.

* histogram.contents.todense() -- this assumes that histogram is sparse and will fail if it is not.  We can change this usage to 'histogram.to_dense(copy=False).contents', which will behave identically if histogram is sparse and will still work if it is dense.

A couple of remaining todense() usages were left as-is, where they densify small subarrays extracted from large sparse Histograms and could incur a serious performance penalty if changed, e.g., histogram[s:e].todense().  This could be made robust to dense vs sparse inputs by using Histogram.slice() instead, but I'll leave that for a future code review of sparse usage in general.

There are no uses of to_sparse/tosparse in cosipy, so there was no need to perform the corresponding fix for this method.

I did not see any failures with the self tests or the notebooks after this change.

Note: this patch does not add any new lines to any file, so deficiencies in patch coverage reflect deficiencies in the original codebase.


israelmcmc:
@jdbuhler Thanks. Looks good, but there are conflicts, I suppose it's with the PR I just merged (#570).

PS: Your IDE really likes to add whitespaces

jdbuhler:
The conflicts are due to using todense() vs to_dense(), which are aliases.  I'll push a fix as soon as I make sure everything is still working.

Re:whitespace, it's actually *removing* extra whitespace at ends of lines.  I blame another project I am involved with that would not take contributions unless they were normalized in this way :-p.

jdbuhler:
Conflicts should be resolved now.

israelmcmc:
Thanks @jdbuhler . Merging!

PR #578, opened 2026-04-17
Don't try to start more Numba threads than the system allows in ts_map

jdbuhler:
@ckarwin noticed that the ts_map tutorial was failing on machines with < 8 cores.  The reason is that fast_ts_fit and moc_ts_fit both suggest using 8 cores as the default (more cores are not worth it in my testing, as there is not enough parallelism available), but Numba is not smart enough to automatically cap the number of threads at the number of available cores.

This patch fixes ts_map to ask Numba how many cores it is allowed to use and manually cap the argument to set_num_threads at this value.


ckarwin (approved):
I successfully ran the ts_map tutorial from the automatic pipeline using 1 core. This PR can be merged.

PR #580, opened 2026-04-17
Update TS map and add transient background tutorials

Yong2Sheng:
## Descriptions
In the generated HTML documentation, the link to the TS map tutorial was outdated. In addition, the transient background tutorial notebook had been added to the tutorials directory, but it was not included in the rendered HTML documentation because it was missing from the tutorial index (`index.rst`). 

## What is changed
- Updated the TS map tutorial link in the tutorial index file to use the current notebook name.
- Added the transient background tutorial notebook to the index file.
- Adjusted the heading structure in the transient background tutorial so the top-level title starts with #, matching the expected document title structure used by Sphinx

### DIFF
--- a/docs/tutorials/background_estimation/transient_background/Transient_background_example.ipynb
+++ b/docs/tutorials/background_estimation/transient_background/Transient_background_example.ipynb
@@ -7,11 +7,11 @@
     "tags": []
    },
    "source": [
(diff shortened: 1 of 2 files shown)


israelmcmc:
Thanks @Yong2Sheng 

Also merged with https://github.com/cositools/cosipy/tree/v0.4.2_rdt

Issue #577, opened 2026-04-17
some tutorials still failing in automatic pipeline


ckarwin:
When running the tutorials from the automatic pipeline, I get three that are failing: 
**spectral_fit_normalizing_flows
ts_map
test_must_fail_magic**

@scipascal @israelmcmc For spectral_fit_normalizing_flows, did something change recently? I get an import error at the very top of the notebook:
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[1], line 14
     10 
     11 from threeML import Band, PointSource, Model, JointLikelihood, DataList
     12 from astromodels import Parameter
     13 
---> 14 from cosipy.threeml.unbinned_model_folding import CachedUnbinnedThreeMLModelFolding
     15 from cosipy.statistics import UnbinnedLikelihood
     16 from cosipy.interfaces import ThreeMLPluginInterface
     17 from cosipy.interfaces.expectation_interface import SumExpectationDensity

ImportError: cannot import name 'CachedUnbinnedThreeMLModelFolding' from 'cosipy.threeml.unbinned_model_folding' (/project/ckarwin/astro/chris/my_envs/cosipy_dev/lib/python3.12/site-packages/cosipy/threeml/unbinned_model_folding.py)
```

@Yong2Sheng For ts_map I get the error below. It looks like you are specifying 8 cpu cores. Do you think it would be better to check available cores on user's system?
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[15], line 2
      1 #[magic commented out by run_tutorials.py] %%time
----> 2 ts_results = ts.fit(nside = 16, energy_channel = [2,3], 
      3                     spectrum = spectrum, cpu_cores = 8)

File /project/ckarwin/astro/chris/my_envs/cosipy_dev/lib/python3.12/site-packages/cosipy/ts_map/fast_ts_fit.py:307, in FastTSMap.fit(self, nside, spectrum, energy_channel, cpu_cores, max_cache_size)
    279 """
    280 Produce a ts map of specified resolution.
    281 
   (...)    303 
    304 """
    306 if cpu_cores is not None:
--> 307     numba.set_num_threads(cpu_cores)
    309 data_cds_array, bkg_model_cds_array, psr_cache = \
    310     self._prepare_inputs(energy_channel, spectrum, max_cache_size)
    312 if self._cds_frame == Frame.LOCAL:
    313     # compute possible source dirs in same frame
[...]

PR #581, opened 2026-04-17
Recommend editable install of ml packages when installing from source

ags-thomas:
This is a 2 insert PR that changes the optional ml package install docs, inspired by discussion in #577 

Currently the docs recommend that users installing optional ml packages from source use

`pip install '.[ml]'`

Most users who install from source will likely be upgrading by git pulling new releases. We already recommend using `-e` when installing the rest of the library from source. When users install the optional packages from source we should recommend they also use the `-e` flag.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -48,9 +48,9 @@ specify the ``[ml]`` extra packages during the installation. e.g.::
 
     pip install cosipy[ml]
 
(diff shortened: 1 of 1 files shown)


israelmcmc:
Sounds good. Thanks @augustus-thomas

PR #583, opened 2026-04-19
Import and comment cleanup

jdbuhler:
While trying to trace the uses of the BinnedData class through the codebase, I had some trouble with false positive hits due to unnecessary imports.  I ended up doing a more thorough cleanup of imports, mainly in the newer code introduced by the interfaces merge, and fixing some comment formatting along the way.

There should be no functionality change after this patch -- just fewer imports in a number of files.


israelmcmc:
Thanks @jdbuhler . The changes look good visually and I'll take the passing test to mean that the behavior didn't change :). Merging!

PR #585, opened 2026-04-20
Fix the IRFAdaptive folding class currently calling the private density and effective area methods.

scipascal:
Thanks to @Yong2Sheng for discovering this! Currently `UnbinnedThreeMLPointSourceResponseIRFAdaptive` calls the private `_event_probability` and `_effective_area_cm2` methods and not the public ones. This results in conflics with response implementations other than mine. In addition:
- I increased the background parameter limits in the tutorial notebook to reduce thrown away samples during fitting
- There was a bug in `update_response_worker_settings` which prevented changing the effective area` batch_size` and `compile_mode` beyond the initialization
- I modified the tests accordingly


israelmcmc:
Thanks @scipascal and @Yong2Sheng . Looks good, I'm merging it.

PR #587, opened 2026-04-20
CHG: Add example on how to install optional packages

zoglauer:
Mini output string change to make it obvious how to fix the error message:
Install cosipy with [ml] optional packages to use these features.
-->
Install cosipy with [ml] optional packages to use these features, e.g., \"pip install cosipy[ml]\".


israelmcmc:
Thanks @zoglauer. Merging.

PR #591, opened 2026-04-21
Further cleanup of to_dense/todense operations

jdbuhler:
Following my previous PR related to to_dense(), this patch reviews the uses of to_dense() and todense() in the cosipy code base (not the tutorials) and implements a few cleanups.  Two notable ones:

* There is one remaining use of .todense() on a Histogram's contents in line background estimation, which assumes that the Histogram is always sparse.  I've patched it to work with either dense or sparse.
* For Histograms arising as PSRs from a FullDetectorResponse, we know that they are always dense.

I've put off addressing use of to_dense() in the tutorials for a later pass, but the majority of such uses are excessive -- the input to the analysis should be densified, rather than having to repeatedly densify its output.


israelmcmc:
Looks good. Thanks @jdbuhler

PR #593, opened 2026-04-28
Add ARMSelector event selection class

Mustafa-hub-maker:
This PR adds a first `ARMSelector` implementation for COSIpy event selection.

The selector applies an ARM cut using:

`ARM = angular separation(source direction, event axis) - Phi`

Updated implementation uses:

- `Chi galactic`
- `Psi galactic`
- `Phi`
- all stored in radians

Validated locally on `GRB_bn110605183`:

- ARM cut: ±13 deg
- GRB total events: 2,776
- GRB selected events: 554
- GRB selected fraction: 0.1996
- ARM distribution is centered close to 0°, as expected

Tests were updated accordingly:

- ARMSelector tests: 5/5 passed locally
- Related event-selection tests: 10/10 passed locally

Note:
This is the validated FITS-column implementation. A remaining refinement is adapting it to the full `ComptonDataSpaceInSCFrameEventInterface`, where `SpacecraftHistory` is used internally for the sky-to-spacecraft transformation.

### DIFF
--- a/cosipy/event_selection/__init__.py
+++ b/cosipy/event_selection/__init__.py
@@ -1,1 +1,3 @@
 from .good_time_interval import GoodTimeInterval
+
+from .arm_selection import ARMSelector
--- a/cosipy/event_selection/arm_selection.py
+++ b/cosipy/event_selection/arm_selection.py
@@ -1,12 +1,26 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[1]:
+
+
 import logging
 
 logger = logging.getLogger(__name__)
 
+from typing import Iterable
+
 import numpy as np
 import astropy.units as u
-from astropy.coordinates import SkyCoord
+from astropy.coordinates import Angle, SkyCoord
+from scoords import SpacecraftFrame
 
+from cosipy.interfaces.data_interface import (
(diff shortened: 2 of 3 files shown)


israelmcmc:
Thanks for working on this @Mustafa-hub-maker.

It's on the right path, but the ARMSelector is not yet working as an `EvenSelectionInterface` implementation. For that to work, `_select` needs to accept an `EvenDataInterface` object, and use it for the calculation. Right now, it's taking a "table-like" object, which doesn't conform with the `EvenDataInterface` protocol definition. Also, the `__init__` method shouldn't take the chi/psi/phi-col parameters, those are also obtained through the input `EvenDataInterface` object in `_select`. You can use existing [`TimeSelector`](https://github.com/cositools/cosipy/blob/develop/cosipy/event_selection/time_selection.py) class as a model for what to do.

On a separate note, it would be good if you use the existing [`RelativeCDSCoordinates`](https://github.com/cositools/cosipy/blob/develop/cosipy/response/relative_coordinates.py) class to compute the ARM. If you feel that your code improves upon what's already available, feel free to modify  `RelativeCDSCoordinates`. I just want to avoid duplicated code if possible.

PR #603, opened 2026-05-06
docs: suggest clone over https rather than ssh

### LINKED ISSUE #589: Change repo cloning recommendation to https
While ssh is in general preferred (more secure, less password handling), cosipy is a public repository so for the the vast majority of user ssh will bring little benefit. However, I've seen many people getting stuck on this first step.

augustus-thomas:
Resolves https://github.com/cositools/cosipy/issues/589

"`git clone`" was found twice in the docs :))

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -32,7 +32,7 @@ accommodate new features in cosipy.
   
 Do the following (preferably inside a conda environment)::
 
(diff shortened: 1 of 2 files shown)


israelmcmc:
Thanks @augustus-thomas !

PR #605, opened 2026-05-08
Cosipy asi pipeline

ldigesu:
In this draft PR, I add the command-line app cosi-tsdetect to the available pipeline tasks. This automates the TS map fitting and extracts the coordinates at the maximum TS value.

To demonstrate the usage I added in docs/tutorials/apps the file run_pipeline_zero.sh, which used all the available apps for an end-to-end analysis pipeline of a point-like source with known tstart-tstop. This is what is currently being used by Francesco Fenu in the prototype of the comprehensive pipeline in cosiflow. I updated the documentation accordingly apps.rst accordingly with all the explanations. I also updated the unit test for the app. I think this PR is of interest for people working in the pipeline @NicoloParmiggiani @falric05 @ckierans 

I still have the following minor problems @israelmcmc :
-for the tutorial pipeline_zero: the orientation file DC3_final_530km_3_month_with_slew_1sbins_GalacticEarth_SAA.fits has either changed name or location in wasabi, and currently the fetching command throws an error

-for the unit test: the apps are not working with the files already available in test_data. However, the test is passed anyway

Once these are solved, this can be converted to a normal PR. I am also open to feedback if things are better arranged in another way.


jdbuhler:
The orientation file you want is now in COSI-SMEX/DC4/Data/Orientation.  The name was changed to start with DC4 instead of DC3 but is otherwise the same.

COSI-SMEX/DC4/Data/Orientation/DC4_final_530km_3_month_with_slew_1sbins_GalacticEarth_SAA.fits

israelmcmc:
Thanks @jdbuhler. I think that solves the first of @ldigesu's points.

As for the other:
> -for the unit test: the apps are not working with the files already available in test_data. However, the test is passed anyway

@ldigesu Can you please clarify what you meant? Is it that the result don't make sense but the unit test still pass? If so, maybe it's due to the test files being very poor --i.e. to coarse. It's enough to test the mechanics, so maybe that's fine for now if you understand what's going on under the hood of the test.

ldigesu:
> Thanks @jdbuhler. I think that solves the first of @ldigesu's points.
> 
> As for the other:
> 
> > -for the unit test: the apps are not working with the files already available in test_data. However, the test is passed anyway
> 
> @ldigesu Can you please clarify what you meant? Is it that the result don't make sense but the unit test still pass? If so, maybe it's due to the test files being very poor --i.e. to coarse. It's enough to test the mechanics, so maybe that's fine for now if you understand what's going on under the hood of the test.

@israelmcmc actually, this specific problem is solved, as we made initial fixes to the test with ASI colleague Marco Pietrosanto. I think tests can still be improved, but this can be done during the review. What is the status of this review, is anyone available? I think Yong or Chris or Carolyne could be interested in reviewing this. Thanks, Laura

israelmcmc:
Thanks @ldigesu. The reviewers you suggested (@Yong2Sheng, @ckarwin or @ckierans) sound good to me. Otherwise I can take a look, but maybe in a couple of weeks.

ldigesu:
Hello, in my last commit to this PR, for pipeline purposes, I modified the plot method in FastTsmap to avoid having the markers l=0 b=0 shown by default in the plots. These are all requests that emerged in the process of committing the comprehensive pipeline to cosiflow.  @Yong2Sheng @israelmcmc, let me know if this is ok. Any news about this review? tks,

Yong2Sheng:
Hi @ldigesu, thank you for the updates. I will take a look at this PR.

Yong2Sheng:
Hi @ldigesu, do you mean the orientation file (`DC3_final_530km_3_month_with_slew_1sbins_GalacticEarth_SAA.fits`) in the TS map tutorial notebook is missing?

Yong2Sheng:
Hi @ldigesu , thank you for putting this together. I focused mainly on the scientific behavior of the TS-map step and its interface with the pipeline. I have a few suggestions and questions that I think would make the task more robust and easier to validate and reuse.

## 1. Time selection, livetime, and background normalization

The current code scales the background using the ratio of the source and background durations. I think the normalization should instead account for the actual [livetime](https://github.com/cositools/cosipy/blob/48ce96ec4e9772a548f941e8f0fb7d4aaf77fa9d/cosipy/spacecraftfile/spacecraft_file.py#L136) recorded in the orientation file, both for the selected GRB interval and for the interval used to construct the background. The duration calculated from `t_min` and `t_max` can give the wrong normalization when there are gaps, dead time, or other excluded intervals.

Related to this, could we define a clearer contract for the background input? It may be more flexible for `cosi_tsdetect` to accept a background model that is already prepared for the selected observation interval, rather than constructing and rescaling it internally. I think we will have different background-estimation methods in the future (please correct me if I am wrong @israelmcmc ), and each method may have different normalization and time-dependence requirements. The method we have in `cosi_tsdetect` looks good now, but we can thinking about how to leave an API for different background modeling methods in the future.

## 2. Energy-channel selection

The TS-map call currently hard-codes:

```python
energy_channel=[2, 3]
```

This selects only one measured-energy bin and may not be optimal for every source or spectrum. Could `energy_channel` be exposed in the configuration file?

In the longer term, it may be useful to select channels based on their expected signal-to-noise contribution, for example by using only channels above a S/N threshold (configurable). That selection logic probably belongs in the TS-map module rather than in the command-line wrapper. For this PR, making the energy channel configurable would already be good.

## 3. CPU configuration

The number of CPU cores is currently fixed to 8. This may fail or behave poorly on machines with fewer available cores, while underusing the capability of machines with more than 8 CPU cores. Could this be configurable, with a default such as `None` (TS map will `use max_cpus - 1` cores if you set CPU cores to`None`)?

## 4. MOC refinement strategy

The default padded MOC strategy uses a very high containment threshold and can produce a map that is close to full resolution, reducing the performance benefit of the multi-resolution approach. I suggest using the initial `nside`, refinement strategy, and containment level in the configuration.

For example, a less aggressive refinement could be tested with:

```python
strategy = MOCTSMap.PaddingStrategy(
    MOCTSMap.ContainmentStrategy(0.92)
)

moc_ts, moc_uniq = ts_map.fit(
    max_nside=8,
    init_nside=1,
    energy_channel=[0, 5],
    spectrum=spectrum,
    cpu_cores=cpu_cores,
    strategy=strategy,
)
```

The exact containment value can be configurable based on the specific purpose.

## 5. Scientific outputs and localization uncertainty

At present, the fitting results are printed to `stdout`. Could these results also be saved in a structured, machine-readable text file for future reference? The file should include at least `max_ts, max_coo, max_l, max_b, pixel_mean_spacing` we get from `get_ts_results`.

The MOC TS map should also be saved for later analysis and reproducibility. You can use the `write_map` method of `mhealpy.HealpixMap`.

Could you also clarify whether the task is intended to report a localization confidence region, such as an ellipse center with its major and minor axes?

## 6. Output overwrite behavior

`cosi-tsdetect` defines an `--overwrite` option, but the variable is not currently used in the code.

## 7. Unit test

The current `test_run_task` only checks that `raw_ts.png` exists. I think we should also check if max_ts, max_coo, max_l, max_b, pixel_mean_spacing are reproduced. In addition, the unit test returns a null map rather than a source-localization validation. Is this expected?
Snipaste_2026-08-19_16-54-23

Have you run the `cosi-tsdetect` with the DC3/DC4 data? It will be a good check to see if the pipeline behaviors properly with simulations.

## 8. Explicit imports

Could we replace the wildcard imports in `task.py` with explicit imports? The wildcard imports obscure the module's actual dependencies and can cause unrelated import changes in the helper modules.

Please import each used symbol directly from its defining module, for example:

```python
from cosipy.pipeline.src.fitting import (
    get_fit_fluxes,
    get_fit_par,
    get_fit_results,
    get_ts_results,
)
from cosipy.pipeline.src.io import load_binned_data, load_ori, tslice_binned_data
from cosipy.pipeline.src.plotting import plot_fit
from cosipy.pipeline.src.preprocessing import get_binned_data, write_yaml
```

Please let me know if you have any questions on my comments. Thanks again for developing this asi pipeline!

ldigesu:
Hello @Yong2Sheng, thanks for your review!  I committed fixes for your points 2-3-4-5-6-8. *About 1*: I agree with you that eventually we will need to input an estimated background here, rather than rescaling the simulated background internally. However, for now, this is the solution that works across all steps of the pipeline. Is the estimated background also working in the spectral fit now? *About 5*: Yes, I think that this task should return the source position as the ellipse center with its major and minor axes. I used the length of the pixel as a rough estimation of the uncertainty, but obviously this is not a proper estimation. Could you share with me the code to obtain the ellipse estimation? I think this could be added to get_ts_results function. *About5*: For now, I would keep the outputs printed to stdout. They are anyway easy to be redirected in a file, using either the command line or a Python script. Eventually, we will have a prescription to follow for the format of the outputs; no need to decide a standard now *About 7*: For now, I make a minimal test, using the test files that are already available, just a check that the tasks run through the end. The tasks themselves are wrappers of functionalities that are already tested by other modules. I'll think about how to make this more robust. Thanks and cheers,

Yong2Sheng:
Hi @ldigesu , thank you for updating the code!
> Is the estimated background also working in the spectral fit now?

Yes, what we can do now is using the 3-month background or the background before and after the burst. It should also work for the spectral fitting.

> Could you share with me the code to obtain the ellipse estimation? I think this could be added to get_ts_results function.

The code I used to obtain the ellipse estimation is not in `cosipy` yet. I presented it in a cosipy meeting, but they thought it was not necessary.... Let me ask Israel first if we should put the ellipse code into cosipy. For unit test, I think checking the coordinate of the peak pixel should be enough.


# Some additional comments:
## `mark_center` will change the current default behavior

This new parameter will affect not only `cosi-tsdetect` but also all calls to `FastTSMap.plot_ts()`. The default of `mark_center` is set to `False`, which will change the default map we have now. Moreover, the position of the parameter will break previous positional calls. For example, if someone used the function like:

```python
plot_ts(ts_map, source_coord, 0.9)
```

then `0.9` will be passed to `mark_center` instead of `containment`, which will break the working script. We should avoid a backward-incompatible API change. A better way to add this new parameter is to use a keyword-only parameter, such as:

```python
def plot_ts(
    m_ts,
    skycoord=None,
    containment=None,
    scheme="nested",
    save_plot=False,
    save_dir="",
    save_name="ts_map.png",
    dpi=300,
    *,
    mark_center=True,
):
```
The pipeline can then explicitly pass `mark_center=False` without changing the default behavior for other users.
Could you please also add `mark_center` to the docstring?

## Use livetime for background modeling

The current background duration is calculated by:

```python
bkg_full = Histogram.open(bk_data_path)
bkg_times = bkg_full.axes["Time"].edges.value
bkg_full_duration = np.ptp(bkg_times)  # max - min
```

However, the orientation contains livetime information, which means there will be periods when the detector is not collecting data. Therefore, the `bkg_full_duration` obtained here will be longer than the actual background livetime. We should count only the time when the satellite is on:

```python
ori_bkg = SpacecraftHistory.open(bkg_orientation)
bkg_full_livetime = ori_bkg.cumulative_livetime().to_value("s")
```

## Explicit imports
The following imports are still not explicit in `task.py`:
```from cosipy.pipeline.src.io import *
from cosipy.pipeline.src.preprocessing import *
from cosipy.pipeline.src.fitting import *
from cosipy.pipeline.src.plotting import *
```

## The overwrite check for the FITS map

The overwrite check for `map_filename` currently checks `plot_filename` again:

```python
map_filename = odir / map_name
if plot_filename.exists() and not args.overwrite:
```

The second condition should check `map_filename.exists()`. I
Additionally, even if you specify `overwrite` here, `get_ts_results()` will still force the map to be overwritten:
```python
m.write_map(map_filename, overwrite=True)
```

## `skycoord=max_coo`

```python
moc_ts.plot_ts(moc_results, moc_uniq, skycoord=max_coo, save_dir=odir, save_plot=True, save_name=plot_name)
```

The `skycoord` parameter in `plot_ts` is defined as the true position of the source rather than the TS peak coordinate. It will not break anything, but I just want to double-check whether you mean to plot the true position or the TS peak position.


Please let me know if you have any additional questions or concerns, thank you again!

israelmcmc:
Thanks @ldigesu and @Yong2Sheng for working on this. 

> The code I used to obtain the ellipse estimation is not in cosipy yet. I presented it in a cosipy meeting, but they thought it was not necessary.... Let me ask Israel first if we should put the ellipse code into cosipy. For unit test, I think checking the coordinate of the peak pixel should be enough.

We can certainly add it to cosipy. I remember your presentation but not exactly how the conversation went. However, while the ellipse is not _necessary_, since you can get the information you want from the map iself, it it certainly useful. The documentation needs to include some caveats, though, since it will not apply to all cases --e.g. the contour doesn't need to be ellipsoidal, and some even have peak in different parts of the sky due to reconstruction ambiguities. 


Btw, I've been trying to follow the conversation, but there's a lot! 😅 Please tag me or let me know if there's something else you need my input for at the moment.

ldigesu:
@Yong2Sheng @israelmcmc 
Hi, this commit addresses Yong's last comments and restructures the unit test, which now includes checks on the results (determined coordinates and flux). I think that if there are no urgent comments, this can be merged, so that at the cosifest the pipelines can be tested with the official cosipy, not with my personal fork. The YS code that computes the uncertainty ellipse can be added to cosi-tsdetect later on, I think. 

@israelmcmc Now all the tests run without errors as you see here; however, despite this, after every commit I am still getting via mail several weird error notifications by codecov....they all seem to be related to a missing token with this kind of error message: 

==> Uploader SHASUM verified (ffb2b821551076b69d3267508b2d5dc10925c821c436521f2216f5ead62cb138  codecov)
==> Running version latest
Could not pull latest version information: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
==> Running git config --global --add safe.directory /Users/runner/work/cosipy_lightcurves/cosipy_lightcurves
/opt/homebrew/bin/git config --global --add safe.directory /Users/runner/work/cosipy_lightcurves/cosipy_lightcurves
==> Running command '/Users/runner/work/_actions/codecov/codecov-action/v4/dist/codecov create-commit'
/Users/runner/work/_actions/codecov/codecov-action/v4/dist/codecov create-commit --git-service github -Z
info - 2026-09-21 14:52:23,465 -- ci service found: github-actions
warning - 2026-09-21 14:52:23,514 -- Branch `cosipy_asi_pipeline` is protected but no token was provided
warning - 2026-09-21 14:52:23,514 -- For information on Codecov upload tokens, see https://docs.codecov.com/docs/codecov-tokens
info - 2026-09-21 14:52:23,820 -- Commit creating complete
error - 2026-09-21 14:52:23,820 -- Commit creating failed: {"message":"Token required - not valid tokenless upload"}
Error: Codecov: Failed to properly create commit: The process '/Users/runner/work/_actions/codecov/codecov-action/v4/dist/codecov' failed with exit code 1

Have you seen something like this before? Is this critical?

israelmcmc:
Thanks @ldigesu. I think those codecov errors are benign. They happen when codecov tried to run on you fork, which doesn't have the permission keys, but the coverage seems fine on this PR which is the one that matter.

I'll let @Yong2Sheng take another look and merge it if he's OK with the changes. On my side, the only thing I saw was that setup.py is still in the changed files, but we delete that one in favor of the .toml a while ago. Can you please remove that file from the branch in this PR?

PR #596, opened 2026-05-01
Modifications to BinnedData.py in DataIO

greendw:
In the DataIO tutorial (https://github.com/cositools/cosipy/blob/develop/docs/tutorials/DataIO/DataIO_example.ipynb),

`analysis.get_raw_spectrum(show_plots=True)`: Now the X,Y axis ranges appear properly.

`analysis.get_raw_spectrum(output_name)`: Now the output file include X-axis errorbars.

A typo correction: 'incidenct' to 'incident'

Examples 3 and 4 have the same file names in `plt.savefig("NAME")`. So, a prefix added to the Example 4 names.

Issue #592, opened 2026-04-27
Outdated link and broken link in DataIO_example.ipynb


israelmcmc:
Cross-reference from https://github.com/cositools/cosi-data-challenges/issues/101 by @jtomsick 

This is the file with the broken links:
https://github.com/cositools/cosipy/blob/develop/docs/tutorials/DataIO/DataIO_example.ipynb

ags-thomas:
This issue is solved in the pull requests https://github.com/cositools/cosi-data-challenge-2/pull/11 and https://github.com/cositools/cosi-data-challenges/pull/109

ckarwin:
I think this can be closed now. Thanks again @augustus-thomas!

PR #607, opened 2026-05-13
Update dataIO tutorial

### LINKED ISSUE #597: Healpix map in Example 2 (DataIO tutorial)
There are two healpix maps in Example 2: upper map in the 'local' and lower map in the 'Galactic' systems.

The Galactic map is fine, but I found that the local map looks very different from the tutorial plot (see attached).

Image

This was generated with the following line:
`analysis.get_binned_data(unbinned_data="unbinned_data.hdf5", output_name="binned_data", make_binning_plots=True, show_plots=True, psichi_binning="local")`

ckarwin:
I updated the dataIO tutorial to be compatible with the latest version of cosipy. Specifically, some of the outputs from the old tutorial didn't match the current cosipy outputs, and additionally, I updated some of the text and a few links. 

This PR fixes the plot range in the raw spectrum, as mentioned in PR #596. @greendw, I already commented on this in your PR, but I went ahead and added the fix in this PR. If you can remove that part from your PR, I think it can be merged once my other comments are addressed. 

This PR also fixes the link that @israelmcmc mentioned in issue #592  that I closed already. 

Lastly, I think this PR addresses the issue #597, where the psichi plot in local coordinates doesn't match what was shown in the old tutorial. First, please note that in the updated tutorial, the local psichi plot is no longer being shown. But if I do make the plot, then I get the same result that is shown in #597. 

I think that the discrepancy might be due to the fact that the old plot is not actually in local coordinates. Instead, it seems to be in celestial coordinates. Note that the data is for 2 hrs of crab observations with the ideal Compton sphere, and it uses an ori file. I think the plot we see now is more in line with what we expect for the local coordinate system. 

To try to trace this back, I looked at an older version of cosipy (v0.3.0): https://github.com/cositools/cosipy/blob/v0.3.0/cosipy/data_io/BinnedData.py
I think the confusion was related to an old issue in histpy where the plot was automatically made in celestial coordinates (although I guess it wasn't clear what the coordinate system was), but this issue has since been resolved. 

@israelmcmc What do you think, does that check out? 

It would be nice if we can merge this before the new cosipy release. 

closes #597


israelmcmc:
Thanks @ckarwin 

> I think the confusion was related to an old issue in histpy where the plot was automatically made in celestial coordinates (although I guess it wasn't clear what the coordinate system was), but this issue has since been resolved.
Yes, makes sense, it was very likely this.

The changes and the preliminary RDT website look good. I'm merging.

@greendw After this merge, it's likely that you'll see a merging conflict in #596, but I think it'll be easy to resolve. Let me know if you have questions.

PR #610, opened 2026-05-19
fix a bug in sc_history

hiyoneda:
I noticed a small error in SpacecraftHistory. It's an important class, and the change is minor, so it should be easy to merge quickly.


israelmcmc:
Thank you, @hiyoneda. Merging.

PR #611, opened 2026-05-19
Remove dependencies on develop wasabi and git trees for files that are used by 0.4.x release

jdbuhler:
The tree COSI-SMEX/develop on wasabi was created during DC4 development to hold new and reformatted data files used by cosipy.  My expectation was that, when cosipy 0.4 was released, the new files in develop would be relocated to the DC4 tree, and no cosipy release code would refer to COSI-SMEX/develop.  This way, we could blow away the develop tree on wasabi and restart filling it with new/reformatted files for DC5.  (In particular, I was planning to upload new responses with stored polarization convention info, per PR #518 ).

However, when DC4 was released, some but not all of the files on which cosipy 0.4 depends got moved.  In particular, we have tutorials and documentation in the 0.4.x releases that depend on the following files from the develop tree that are not part of DC4:

* Data/Orientation/20280301_3_month_with_orbital_info.fits

This is an orientation file that was used in DC3 but has been reformatted to FITS format.

* Data/Responses/SMEXv12.Continuum.HEALPixO3_10bins_log_flat.binnedimaging.imagingresponse.h5
* Data/Responses/SMEXv12.511keV.HEALPixO4.binnedimaging.imagingresponse.h5

These are old responses used by cosipy tutorials; they were reformatted vs DC3 (so we shouldn't just overwrite the corresponding files in the DC3 tree) but have been superseded by newly generated (and presumably different) responses in DC4.

* Data/Responses/PointSourceResponse/psr_gal_511_DC2.h5
* Data/Responses/PointSourceResponse/psr_gal_511_DC2_earthocc.h5
* Data/Responses/PointSourceResponse/psr_gal_continuum_DC2_earthocc.zip

These are PSRs from the DC2 tree that have been reformatted to remove overflow bins and (I think) to use internal compression.

Additionally, there are some other files in develop (FITS files beginning with Data/Orientation/DC3_final_530km_3_month_with_slew_....) that are directly referenced by the docs and tutorials of 0.4.x.

This patch removes all dependencies of the tutorials and documentation on COSI-SMEX/develop.  After conversations with @ckarwin, we determined that

* The DC3_final_... orientation files are present in DC4, but they were renamed "DC4_final...".
* The reformatted 2028... orientation file could be moved to the DC3 directory, since its extension is distinct from the .ori file already there.
* The remaining response files could be moved to the existing COSI-SMEX/cosipy_tutorials tree on wasabi.  I created a directory structure under there (Data/Responses) that mirrors the structure present in the develop directory, and Chris copied the files.

After copying the files out of COSI-SMEX/develop as described, this patch redirects all references to them to point to their new locations.  Additionally, references to the old .ori versions of the orientation files that slipped into some tutorials at the last minute were updated to point to the newer .fits versions.  The patched repo passes the run-tutorials script (though my local install doesn't run the ml-dependent code).

A related issue is that two URLs in the API and tutorial docs point to tutorials in the github develop branch. These have been rewritten to point to the main branch (which was already the case for other URLs in the same documentation).

Ideally, this patch would be applied both to develop and to the release branch of cosipy.  A remaining question is, when can we safely assume that the files currently in COSI-SMEX/develop are no longer in use and can be overwritten or deleted?  Presumably, this can't be done immediately, as the current 0.4.3 release points to them.  But if we release 0.4.4 with this patch, can we then break the older 0.4 releases by changing the develop tree?


ckarwin (approved):
Thanks @jdbuhler. I successfully ran all the tutorials. This can be merged now.

israelmcmc:
Thanks @jdbuhler and @ckarwin. I cherry picked @jdbuhler latest commits and merged then into the v0.4.3 documentation branch (#612). I should take a few minutes to get updated in ReadTheDocs

PR #616, opened 2026-05-27
Use response file polarization conventions in tutorials

jdbuhler:
A recent update to cosipy ( PR #518 ) added support for storing polarization conventions inside response files, which is useful for both polarization and relative binned responses.  This patch updates tutorials using the DC3/4 o3 polarization response to expect the response to contain a stored convention, rather than supplying one explicitly on the command line.

I have uploaded an updated pol response to COSI-SMEX/develop/Data/Responses, and the tutorials that need a pol response now use this one.  I also fixed the docstring for the source injector to reflect that a convention is needed only if the file doesn't have one.

Note that this is for develop *only*, since the relevant PR was merged after the 0.4 release.

### DIFF
--- a/cosipy/source_injector/source_injector.py
+++ b/cosipy/source_injector/source_injector.py
@@ -10,7 +10,8 @@ logger = logging.getLogger(__name__)
 
 class SourceInjector():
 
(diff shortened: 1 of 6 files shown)


eneights (approved):
Looks good, and the notebooks run successfully with no change in the output

PR #617, opened 2026-05-27
Fix ResponseConverter's renaming of .rsp and .rsp/.gz files

jdbuhler:
When converting an .rsp or .rsp.gz response file to HDF5, the ResponseConverter can automatically choose the HDF5 file's name if none is provided.  The intended behavior is that foo.rsp or foo.rsp.gz becomes foo.h5 .  Unfortunately, the existing name formation code was buggy.  In particular, if the base filename contains additional periods (e.g., foo.bar.rsp), part of the base could be deleted.

This patch fixes the renaming code to avoid this bad behavior.

### DIFF
--- a/cosipy/response/RspConverter.py
+++ b/cosipy/response/RspConverter.py
@@ -238,12 +238,13 @@ class RspConverter():
         if h5_filename is None:
             rsp_path = Path(rsp_filename)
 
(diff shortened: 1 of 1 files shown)


israelmcmc:
Thanks @jdbuhler! Looks good, merging!

PR #620, opened 2026-06-04
Unify ASAD and Stokes method polarization code

jdbuhler:
The ASAD and Stokes method code to determine polarization angle and fraction share a great deal of functionality, though it is implemented differently in each method.  This patch extracts the common functionality from each method into a base class, PolarizationFitting, and derives PolarizationASAD and PolarizationStokes from that.  Besides avoiding a great deal of code duplication, this change lets us use the fast ASAD calculations from the ASAD method in the Stokes method as well, which reduces the latter's overall running time to a fraction of a second for GRB-sized inputs and 2-3 minutes to analyze 3 months of simulated events + background from CygX1.

The common code is mainly in computing scattering angles from data/background and in computing ASADs as input to the mu100 and mdp99 calculations. The actual fitting code is, of course, quite different and is implemented separately in each subclass.  The other key differences are that the Stokes code supports calculation in the absence of background data, which the ASAD code does not, and that the ASAD method supports binned input data, which the Stokes code does not.  Originally, the Stokes code used a different implementation of the mu100 calculation that produced slightly different answers, but after consulting with @eneights and @nmik , I used the ASAD code's implementation for both methods. (This can easily be changed if desired.)

The Stokes fitting code has been updated to follow a newer reference (Kislat and Spooner 2023), with errata corrections as directed by @nmik, and to compute background-subtracted normalized Q and U values.  Thanks to @mabp1992 for independently implementing Stokes fitting; reconciling that code with this code smoked out several issues on both sides.

The ASAD and Stokes notebooks have been updated to the API of the new implementation, which involved renaming a couple of properties and extracting some plotting code for Stokes into its own function.  The notebook versions on this branch show the output from each method under the unified code.

To improve test coverage, the same two unbinned test cases used for the ASAD method have been added to the Stokes method test suite, in addition to the existing no-background test.  These tests raise numerical errors from numpy due to square roots of negative numbers, but the current belief is that these errors stem from the very coarse response binning (nside 1) in the test cases, rather than any actual error.


jdbuhler:
Taking this out of draft status... I think it should be reviewed now, even if we decide some changes are needed.

israelmcmc:
Thanks @jdbuhler! Let me find a reviewer.

eneights:
The ASAD code looks good to me and the notebook outputs the same results as before. The Stokes method notebook now has slightly different results, so @nmik should take a look to make sure the changes are okay

PR #621, opened 2026-06-08
Code cleanups and minor performance improvements

jdbuhler:
This patch collects numerous small changes I made while working on other things.  Aside from some comment cleanups, the changes include

 * allow ts_map code to use the FP type of the data for all operations, rather than defaulting to 64-bit; using single-precision can be a significant (on the order of a couple of seconds) performance win
  * take advantage of PolarizationAxis's copy=False argument
  * remove no-op test in FullDetectorResponse.get_point_source_response()
 * don't require string->PolarizationConvention mapping through default function arguments
 * simplify Axes construction in  instrument_response.py
 * force some plotting code to close its figure on completion, to avoid test-case warnings about too many open figures
 * suppress ERFA warnings from time_selector tests

### DIFF
--- a/cosipy/polarization/conventions.py
+++ b/cosipy/polarization/conventions.py
@@ -242,14 +242,18 @@ class MEGAlibRelative(ConventionInSpacecraftFrameMixin, OrthographicConvention):
 
         axis = axis.lower()
 
-        if axis == 'x':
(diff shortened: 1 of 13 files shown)


GallegoSav (approved):

israelmcmc:

PR #624, opened 2026-06-23
Added Xspec capability to cosipy

mabp1992:
This pull request tries to add output unit of the Xspec models and the embedded integrator of the models used during maximum likelihood analysis.

### DIFF
--- a/cosipy/response/functions.py
+++ b/cosipy/response/functions.py
@@ -101,13 +101,19 @@ def get_spectrum_unit(spectrum):
             case Band_Eflux():
                 spectrum_unit = spectrum.K.unit / spectrum.a.unit
             case _:
-                spectrum_unit = None
-                for pname in ('K', 'k'):
-                    if pname in spectrum.parameters:
-                        spectrum_unit = spectrum.parameters[pname].unit
+                # spectrum_unit = None
+                # for pname in ('K', 'k'):
+                #     if pname in spectrum.parameters:
+                #         spectrum_unit = spectrum.parameters[pname].unit
 
-                if spectrum_unit is None:
-                    raise RuntimeError("Spectrum not yet supported because units are unknown.")
+                # if spectrum_unit is None:
+                #     raise RuntimeError("Spectrum not yet supported because units are unknown.")
+                
+                
+                # ====================================================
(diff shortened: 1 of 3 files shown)


cosipy/response/functions.py
```diff
@@ -101,13 +101,19 @@ def get_spectrum_unit(spectrum):
...
                 spectrum_unit = spectrum.K.unit / spectrum.a.unit
             case _:
-                spectrum_unit = None
-                for pname in ('K', 'k'):
-                    if pname in spectrum.parameters:
-                        spectrum_unit = spectrum.parameters[pname].unit
-
-                if spectrum_unit is None:
-                    raise RuntimeError("Spectrum not yet supported because units are unknown.")
+                # spectrum_unit = None
```
GallegoSav, line 104: @mabp1992 please don't comment the case `K,k` because we use it for other source. Just add a specific case related to your XSPEC thing

cosipy/response/integrals.py
```diff
@@ -156,31 +176,87 @@ def get_integral_values(f, x_in, force_quad = False):
...
-    Compute the integral of a function f between the specified
-    endpoints using adaptive quadrature
 
-    Inputs
-    ------
-    f : function of type float -> float
-    x : array of float
-      array of monotonically increasing grid points; integration is
-      performed between each successive pair of points
+# def integral_generic(f, x):
```
GallegoSav: Also here you replace the `integral_generic` method by something else which broke all the unit test. Instead maybe add a method for the specific Xspec spectrum

israelmcmc:
@mabp1992 It seems that the only changes that are actually needed to make Xspect models compatible are the ones in integral.py, is that correct? Those look fine do me. If you undo the changes to FullDetectorResponse.py and function.py it would address @GallegoSav's remaining comment and we can merge this. @GallegoSav do you agree? Or is there's something else still pending?

mabp1992:
No. You need to define the unit of the spectrum in the "None" situation. Also, there is a bug in "repr()" function when you want to see the polarization response file. I fixed that too.

GallegoSav (approved):

israelmcmc-ai:
Thanks @mabp1992 and @GallegoSav

PR #632, opened 2026-07-07
Make SpecFromDat more general

GallegoSav:
The read function of the dat file for `SpecFromDat` was expecting a very particular format in order to work. I made it more general where it will only read the line with `DP` and skip all the rest.


ckarwin (approved):
Thanks, @GallegoSav. This is a good improvement. I tested the new code on a DC4 .dat file and it works ok.

PR #633, opened 2026-07-08
Add custom PointSource and ExtendedSource

GallegoSav:
This pr add a custom threeml PointSource made by @jdbuhler in order to improve significantly computing time during the fit.

This is mostly important for the unbinned analysis where we will use :
`from cosipy.threeml.custom_functions import CosipyPointSource `

instead of : 
`from threeML import PointSource`

Edit : we will do the same for the ExtendedSource in this pr

### DIFF
--- a/cosipy/event_selection/good_time_interval.py
+++ b/cosipy/event_selection/good_time_interval.py
@@ -96,6 +96,60 @@ class GoodTimeInterval():
         self._tstart_list = self._tstart_list[sort_idx]
         self._tstop_list = self._tstop_list[sort_idx]
 
+    def contains(self, times, time_format='unix', time_scale='utc'):
+        """
+        Check which times fall inside this GTI.
+
+        Parameters
+        ----------
+        times : astropy.time.Time or array-like
+            Times to test. Numeric inputs are interpreted with
+            ``time_format`` and ``time_scale``.
+        time_format : str, optional
+            Astropy time format for numeric inputs. Default is 'unix'.
+        time_scale : str, optional
+            Astropy time scale for numeric inputs. Default is 'utc'.
+
+        Returns
+        -------
+        numpy.ndarray or numpy.bool_
+            Boolean mask with the same shape as ``times``. Intervals
+            use the half-open convention ``start <= time < stop``.
+        """
+
+        input_is_time = isinstance(times, Time)
+        if input_is_time:
+            time = times
+        else:
+            time = Time(times, format=time_format, scale=time_scale)
+
+        result_shape = np.shape(time)
+        if len(self) == 0:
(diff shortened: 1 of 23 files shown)


GallegoSav:
@israelmcmc I implemented the optimization made by @jdbuhler for the Point and Extended Source and here is the benchmark results . The best result is when combining this new class with the pytorch implementation of the spectral function.

The best result is a factor 10 for the point source and 3 for the extended source. 

image
image

image
image

israelmcmc:
Cool! Is this implementation tuned for cosipy or is it a general implementation that can works for all instruments?

I'm wondering about this:
- Will this work on a joint analysis?
- Is this a candidate to eventually replace Source/ExtendedSource in 3ML rather than living in cosipy?
- What are the main reasons for this speed gain?

jdbuhler:
I'm trying to track the same changes @GallegoSav is making as Astromodels PR 285.  So yes, if the 3ML folks are good with the changes, they could go into Astromodels instead of living forever in cosipy.

The main benefit seems to arise from the replacement for Python sum(), which is what we hashed out with Pascal back in the spring. I see no reason that this would not work for any use of Astromodels.

israelmcmc:
Thanks @jdbuhler. Sounds good. Tagging @ndilalla and @omodei for awareness. Also here's the link so it gets automatically referenced https://github.com/threeML/astromodels/pull/285

GallegoSav:
@jdbuhler is that possible to merge this pr ?

jdbuhler:
Have not been following this recently - is the review completed?

israelmcmc:
Can we make CosipyPointSource inherit from PointSource instead of Source? And equivalent for the extended version. That way we don't need the rest of the medications like

```
if not (isinstance(source, ExtendedSource) or isinstance(source, CosipyExtendedSource)):
```

Since `isinstance`catches all the inherited classes. This will also help make future code more maintainable, since they won't need this special if case.

Besides that I'm good with the code, since it would be contained to these new two classes. After that, if either @jdbuhler or @scipascal have run it (I haven't) I'd be OK to merge this.

jdbuhler:
I would rather that @scipascal confirm that this works with whatever ML mods we are doing.  I can hack on it tomorrow or Wednesday if @GallegoSav does not get to it first.

GallegoSav:
@jdbuhler @israelmcmc , Ok I made the change + I am using this new classes for both binned and unbinned analysis

israelmcmc-ai:
Thanks @GallegoSav. I'll go ahead and merge this since it's basically a no-harm change. Only the new (optional) classes in custom_functions.py and tests (the rests are now just spaces). Is we find an issue along the way we can open a new PR.

PR #637, opened 2026-07-30
Fix issue 636

### LINKED ISSUE #636: Big-Endian byte order from fits file raise errors for tensors
FITS files are historically written in Big-Endian byte order (>), whereas modern CPU architectures (x86_64 / ARM) use Little-Endian byte order (<).

When astropy.io.fits reads the ori file from a FITS file, it keeps the underlying NumPy array in Big-Endian format. `PyTorch's torch.as_tensor()` cannot process arrays with non-native byte orders and immediately raises:
`ValueError: given numpy array has byte order different from the native byte order.`

So we need to explicitly cast the fits table into `np.ascontiguousarray` in [_open_fits](https://github.com/cositools/cosipy/blob/c0bc11e4d8ae3be8dcc25469abd26c6cbead69e3/cosipy/spacecraftfile/spacecraft_file.py#L315). 

I will open a pr to fix that. This is mainly for the `livetime` array.

GallegoSav:
pr for fxing the issue #636 .

### DIFF
--- a/cosipy/spacecraftfile/spacecraft_file.py
+++ b/cosipy/spacecraftfile/spacecraft_file.py
@@ -334,6 +334,22 @@ class SpacecraftHistory:
 
         t = QTable.read(filename)
 
+        #Convert EVERY column to native byte order ---
+        for colname in t.colnames:
+            col = t[colname]
+            # Check if the column has a unit attached
+            if hasattr(col, 'unit') and col.unit is not None:
+                unit = col.unit
+                # 1. Extract raw numpy array via .value
+                # 2. Make it contiguous & native endian
+                # 3. Re-wrap with original unit
+                native_array = np.ascontiguousarray(col.value)
+                t[colname] = u.Quantity(native_array, unit=unit)
+        
+            # Handle plain numpy columns (no units)
+            elif isinstance(col, np.ndarray):
+                t[colname] = np.ascontiguousarray(col)
+        
         # make sure we have version info, and that we support this version
         if "VERSION" not in t.meta:
             raise ValueError("FITS orientation file has no version info")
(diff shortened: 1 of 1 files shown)


jdbuhler (changes requested):
I understand why the byte order needs to change explicitly, but why do you also have to do ascontiguousarray()?  As far as I can tell, even if an array (or Quantity) is not C-contiguous before it is written to a FITS file, it becomes contiguous when it is read back in.  Example:

x = Quantity(np.asfortranarray([[1,2,3],[4,5,6]]), unit="s")

x.flags
 C_CONTIGUOUS : False
  F_CONTIGUOUS : True
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

t = QTable(data=[x], names=["foo"])
t.write("foo.fits", format="fits")

t = QTable.read("foo.fits")
x = t["foo"]

x.flags
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False

If you don't have to do ascontiguousarray(), then you can simplify your code quite a bit, because astype() preserves Quantity objects.  So you should be able to just say

t[colname] = np.astype(col, col.dtype.newbyteorder('='), copy=False)

whether col is a quantity or a bare array.  (Note the copy=False, which prevents astype from making a copy if the byte order doesn't have to be changed).

GallegoSav:
@jdbuhler thanks for the comment ! I implemented it

jdbuhler (approved):
Looks good!

israelmcmc:
Thanks @GallegoSav and @jdbuhler !

PR #638, opened 2026-08-05
Add Cutoff PowerLaw torch version 

GallegoSav:
Similarly to the pr #628 , this one add an exponential cutoff power law using torch. 

Much faster for evaluation, to be used for unbinned analysis.

### DIFF
--- a/cosipy/threeml/ml/function_torch.py
+++ b/cosipy/threeml/ml/function_torch.py
@@ -62,7 +62,7 @@ class FastPowerlawPyTorch(Function1D, metaclass=FunctionMeta):
         # The normalization has the same units as the y
 
         self.K.unit = y_unit
-        
+        self._y_unit = y_unit
     
     def evaluate(self, x, K, piv, index):
         
@@ -74,7 +74,7 @@ class FastPowerlawPyTorch(Function1D, metaclass=FunctionMeta):
 
         if isinstance(K, astropy_units.Quantity):
             K_, piv_, index_ = K.value, piv.value, index.value
-            unit_ = self.K.unit
+            unit_ = self.y_unit
         else:
             K_, piv_, index_ = K, piv, index
             unit_ = 1.0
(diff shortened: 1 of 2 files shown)


jdbuhler (commented):
Looks pretty much like the code for your original PyTorch powerlaw.  Two questions:

(1) Why does the index parameter default to -2.01 for the regular powerlaw, but -2 for the cutoff powerlaw?

(2) I'm a little confused as to how/if you decide to set y_unit.  In all classes in this file, you grab self.y_unit to decide the unit of the output if the input is a Quantity.  But I don't see that being set anywhere.  The cutoff powerlaw class sets self._y_unit from the y_unit argument in _set_units(), which the other classes do not, but it only reads from self.y_unit, not self._y_unit. Is this some kind of Astromodels magic, and if so, why is it not consistent between classes?

GallegoSav:
@jdbuhler 

(1) I kept the exact definition from threeml so I don't know the reason of the index difference. Maybe that's a typo ?

(2) So actually in the unit test, when it was trying to test the class with Quantity , it raised error because `self.y_unit` (the method from `Function1D` that return `self._y_unit`) was None. 
I see `y_unit` is set by the [method](https://github.com/threeML/astromodels/blob/3515d4ca48dbd3f5fe2e4aa2c1fa5d9988da3e39/astromodels/functions/function.py#L1200) `set_unit` from `Function1D`.  So we actually need to call `set_unit` and then `_set_unit` in the unit test. 

I did the modification and removed the `self.y_unit = y_unit` from the pr

jdbuhler (commented):
I thought the external Function1D interface offers set_units(), which in turn calls _set_units() internally. So you should never have to call _set_units() or directly set ._x_unit or ._y_unit yourself outside the class.

Does the test code not work with just the set_units() call?

GallegoSav:
> I thought the external Function1D interface offers set_units(), which in turn calls _set_units() internally. So you should never have to call _set_units() or directly set ._x_unit or ._y_unit yourself outside the class.
> 
> Does the test code not work with just the set_units() call?

Yes it worked :+1:

jdbuhler:
Great! Please also replace the explicit setting of _xunit and _yunit at in the earlier tests, and then I think you're golden

jdbuhler:
(There is also one remaining call to _set_units that should be updated)

GallegoSav:
@jdbuhler done !

jdbuhler (approved):
Yay!

israelmcmc:

PR #639, opened 2026-08-17
Add super cutoff pl torch version 

GallegoSav:
Similar to previous pull requests, this one add a version of the threeml super cutoff power law using torch for faster evaluation.

### DIFF
--- a/cosipy/threeml/ml/function_torch.py
+++ b/cosipy/threeml/ml/function_torch.py
@@ -149,11 +149,12 @@ class FastGaussianPyTorch(Function1D, metaclass=FunctionMeta):
         
         
         norm = self.__norm_const / sigma
-        x = torch.as_tensor(x, dtype=torch.float64, device=self.devices.value)
+        x_t = torch.as_tensor(x, dtype=torch.float64, device=self.devices.value)
+        x_t = x_t.view(-1, 1)
         mu = torch.as_tensor(mu, dtype=torch.float64, device=self.devices.value)
         sigma = torch.as_tensor(sigma, dtype=torch.float64, device=self.devices.value)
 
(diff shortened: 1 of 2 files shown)


jdbuhler (changes requested):
Your definition of super cutoff powerlaw appears to be 

K * (x/piv)**index * exp(-x/xc)**gamma

But I think the Astromodels definition is

K * (x/piv)**index * exp(-(x/xc)**gamma)

which is not the same.

Also, you might want to consider the roundoff-resistant version of the computation that Astromodels implements:

log_v = index * np.log(x / piv) - np.pow(x / xc, gamma)
return K * np.exp(log_v)

ckarwin:
Thanks for pointing this out @jdbuhler. I noticed the same thing this morning about the wrong model definition.

jdbuhler (commented):
One other question -- in this function, as well as your other Torch implementations, you squash the input tensor x to 2D by linearizing all but the last dimension.  Is there a particular reason for doing this?

GallegoSav:
@jdbuhler thanks for spotting the mistake ! 

I changed it to match the astromodel version. 

The `x_t.view(-1, 1)` is suppose to guarantee clean 2D broadcasting against model parameter tensors (e.g. evaluating $N$ energy points against $M$ spectral parameters or source components simultaneously to produce an $(N, M)$ evaluation matrix). Although this step could be optional since I don't think we will use something else than 1D shape array for the energy.

jdbuhler (approved):
Issues with function definition have been addressed. Looks good

israelmcmc:
Thanks @GallegoSav , @jdbuhler and @ckarwin . I'm merging it.

PR #640, opened 2026-08-17
Warning for MDP > 100%

eneights:
This adds a warning when PolarizationASAD is initialized if the MDP of the source exceeds 100%


israelmcmc:
Thank @eneights. I'm merging it.

PR #642, opened 2026-08-21
Fixes from PR 641 

israelmcmc:
These are localized fixes found while working on #641, but that are not specific to the IRF introduced there.


israelmcmc:
I'm merging this. It passes the tests and it doesn't touch anyone else's code.

PR #644, opened 2026-08-26
General binned sed

ckarwin:
This PR introduces several updates to improve spectral fitting with COSIPy, including a new generalized binned SED model, substantially improved performance for extended-source fitting, and analytical integration for supported spectral models.

### Main changes

* **Generalized binned SED fitting:** Added a new `BinnedSED` spectral model for reconstructing a spectral energy distribution directly in true-energy bins. The spectrum is represented as a piecewise power law, with an independent normalization in each selected energy bin and a shared local spectral index. The model can be constructed for an arbitrary number of contiguous true-energy bins defined directly from the instrument response.

* **Analytical integration of the SED model:** Added an exact analytical integral for `BinnedSED`, avoiding repeated numerical integration when folding the model through the response.

* **Faster extended-source spectral fitting:** Improved the response calculation for extended sources with energy-independent spatial morphologies. For a fixed morphology, the computationally expensive spatial convolution is now performed only once. Subsequent likelihood evaluations reuse the cached spatially convolved response and only refold the changing spectral model over the true-energy axis, substantially reducing the cost of iterative spectral fitting.

* **Analytical integration of `Super_cutoff_powerlaw`:** Added analytical integration for the astromodels `Super_cutoff_powerlaw` model, further reducing the need for repeated numerical integration during response folding.

* **SED fitting utilities:** Added helper functions for configuring and working with the binned SED model, including identifying poorly constrained bins and calculating profile-likelihood upper limits.

* **Tutorial and tests:** Added an example tutorial notebook demonstrating how to use `BinnedSED`, including the fitting workflow and upper-limit calculation. The notebook was added in other_examples and included in the tutorial autorun YAML.

Associated unit tests were also added or updated for the new functionality. All tutorial notebooks were run to verify that they complete successfully with these changes.


ckarwin:
@parshadkp You may be interested in this PR.

GallegoSav:
OK the notebook is working for me , we can merge.

GallegoSav (approved):

PR #648, opened 2026-09-10
Update to BinnedSED to include other astromodels shapes

parshadkp:
PR #644 added a BinnedSED class to produce SED plots, which uses a powerlaw with a frozen index = -2 to determine the flux in each COSI bin. Updated BinnedSED class to use any user specified shape for the SED fitting. This turns out to yield more accurate SED points when the user specified model matches the injected model. The specified model for BinnedSED can be any astromodels. A user can also perform a global fit on the data and pass in the globally fitted spectral model into BinnedSED. 

Providing the correct spectral shape to BinnedSED seems to work better compared to the default powerlaw case. Especially for obtaining better upper limits. Note this is for one realization of many that can be performed, which I will continue to work on. As per discussions with @israelmcmc and other members, the two methods may not be too different if we account for finer bins or background uncertainties etc...

image
image
Screenshot 2026-09-15 at 11 35 02 AM

Core Updates:
- Removed index argument from BinnedSED and added a spectral_shape to pass in the desired shape for the SED fit.
- Added to the unit tests for the updated code.
- Fixed F-string typo in "cosipy/cosipy/threeml/sed_utils.py".

Let me know if additions to the tutorial notebook are necessary for these changes!

@ckarwin could you please take a look when you get the chance? I had to re-run the tutorial notebook because of a small change so let me know if the outputs still make sense.


ckarwin:
@parshadkp Thanks for making these changes. The generalized fixed reference shape looks reasonable to me, as long as only the normalization is fit by default in each bin. I would also like to retain the functionality of the original default power law: the index should be adjustable from the default value of -2, and it would be useful to optionally allow the single shared power-law index to remain free. Right now it looks like all parameters of the supplied spectral shape are always frozen, so could you preserve that flexibility for the default PL?

Separately, while comparing the SED from this PR with the existing tutorial, I noticed that the detected bins in the original tutorial had substantially smaller uncertainties. I looked into this and found that the plotted errors were being taken from Minuit's HESSE covariance estimate, which appears to be numerically unstable for the small, bounded flux parameters in this fit and can significantly underestimate the uncertainties. A more robust approach is to determine the errors directly from the likelihood crossings at $-\Delta\mathrm{logL}=0.5$.

I've attached an updated notebook that implements the likelihood-based errors and also cleans up some of the SED handling. Could you use this version as the starting point and make the changes needed for compatibility with the generalized `BinnedSED` implementation in this PR? Hopefully the resulting SED should then reproduce the same flux points and uncertainties. At the top of the tutorial you might also want to add a few lines mentioning that the tutorial uses the default PL, although any spectral shape can be used. 



[sed_spectral_model_tutorial_replacement.ipynb](https://github.com/user-attachments/files/32068104/sed_spectral_model_tutorial_replacement.ipynb)

parshadkp:
Hi @ckarwin, thanks for looking this over. The powerlaw index can be adjusted by supplying a powerlaw shape to the spectral_shape argument. I will update the class to optionally free the index or any other parameter given user's choice.

I will also look over the updated notebook and add a few lines that you mentioned.

parshadkp:
Hi @ckarwin, I updated binned_sed.py with the capability of freeing individual parameters as needed in the bin of choice (or all). I also ran and updated the tutorial notebook and verified it gives me the same results as the notebook that you attached. https://github.com/cositools/cosipy/pull/648/commits/89bc4b54460697f4b903e765ec1e9d39c599b1e7

I am also removing the draft label now as I added the comparison plots to the original pull request message.

Let me know if you think there is anything else missing.

ckarwin:
Hi Parshad, Thanks for making those changes. They look good. 

Can you upload another comparison plot? The resolution of the current one is very poor and not really visible. Related to this, I think there is an important distinction that needs to be made with the different spectral shapes used for the SED calculation. When plotting the SED, it is best to include the x error bars that span each bin. Actually, the x error bar should show the shape of the spectrum in the bin. Typically, you see SEDs with flat x error bars (when plotting E^2*dN/dE). But actually, this is only true when the index=-2. If one frees the PL index, the x error bars should show the best-fit slop. 

In your case, apparently you are fitting a PL with exponential cutoff for the bin shape. So in the higher energy bins I am not quite sure what to make of the upper limits, because the spectral shape becomes very steep. Looking at your comparison plots, the comparison looks very similar expect for the first case with E_c = 200 keV. Your plot seems to indicate that the yellow data points have better sensitivity at higher energy. Again, I don't really know how to interpret this, because you're not actually fitting a flat spectrum in these bins. Rather, you are imposing a cutoff from the spectral shape.   

I suppose the functionality is fine for this PR. But for any actual analysis that would eventually be published, this distinction would need to be made very clear. In practice, I'm not sure how useful it is to show an "SED" which was calculated with a PL with exponential cutoff spectral shape.

parshadkp:
Hi @ckarwin, thanks for the comments! I updated the plots with better quality ones. Note that these are still ones without x error bars. Changing the x error bars from horizontal to more representative ones is a great idea. I will continue to work on this and I can update you when I finish through slack or in-person.

The idea behind the supplying the spectral shape is that you are performing a global fit to the data first and then using that fitted spectral shape for the sed. This does make the sed more model dependent but may be useful in some cases where you want model consistent flux points. I believe this is not useful for the tutorial since it doesn't concern a physical model but for a source with a cutoff powerlaw or some other combined spectral shape this may be useful. I believe Gammapy uses a similar approach, where users can define a spectral shape for their SEDs.

Your comment about the upper limits is valid and I can look into this in more detail as well. I think the functionality is good to have in cosipy and I can issue a new PR later if needed later to fix any issues. In any case, the default method still works as before and now we have the ability to free the index for given bins.

ckarwin:
Thanks again, @parshadkp. Ok, I think this PR can be merged now.

ckarwin (approved):
Looks good.

israelmcmc:
Thanks @parshadkp and @ckarwin !

Issue #238, opened 2024-08-29
Enhance data selection methods in dataIO class


ckarwin:
Currently, only time cuts are available. This should be extended to allow for other cuts.

Additionally, when we get to the point of doing any kind of unbinned time resolved analysis, something analogous to slice() would be very useful, so we don’t have to create and read in separate files for each time range.

israelmcmc:
I'm working on an EventSelectionInterface in #369. As usual, it'll make this easier to implement this, but well, the specifics still need to be implemented. If someone want to work on this let's coordinate first.

ckarwin:
Hey @israelmcmc I'm interested in working on this. Let's coordinate!

avalluvan:
Issue #361 has a related addition: GallegoSav added Compton Sequence to the unbinned data file and a function to perform event selection over it.

israelmcmc:
With the interfaces refactoring, each variable can be it's own EventSelection. I think we should split this issue into a few more concrete issues, specifying which variable you want a selector for for each issue.

PR #651, opened 2026-09-22
Energy selector

israelmcmc:
I'm adding an EnergySelector, which is an EventSelectorInterface implementation. It can work for multiple energy ranges. You can combined, intersect or exclude energy ranges. 

I originally needed this while working on #641 for fit only an energy range, but it's something that we need in general (e.g. filter out energies when working on the continuum).

Related to #238


scipascal:
@israelmcmc I took a look at your code and it seems all right. Tomorrow I will make the necessary small modifications to my code and run a line fit to test it, then it can be merged.

scipascal (approved):
As a test, works for my multi-line fits as expected. I am merging.

israelmcmc:
Thank you, @scipascal !

